PR #8, opened 2022-06-03
Month method

vasconsaurus:
Started to write the month method for item season, does this make sense?


app/models/item_season.rb
```diff
@@ -9,4 +9,33 @@ class ItemSeason < ApplicationRecord
   belongs_to :produce_item
 
   validates :month_index, :country_code, presence: true
+
+  def month(month_index)
```
vasconsaurus, line 13: @tcione if this makes any sense, I thought it could be simplified by something like this:
```
def month(month_index)
months = %w[january february march april may june july august september october november december]
months[month_index - 1]
end
```

tcione: Generally this is better handled via i18n (internationalization). Luckily Rails does already provide month names in its default strings, so this method could become simply:
```ruby
def month(index) = I18n.t("date.month_names")[index].downcase
``` 

A nice thing is that we can easily translate the app in the future adding something like https://github.com/svenfuchs/rails-i18n

OBS: I've used the [endless method definition](https://www.bigbinary.com/blog/ruby-3-adds-endless-method-definition) introduced in Ruby 3.

tcione (approved):
Codewise everything looks good (thus the approval). However I'm a bit confused about the PR title and description. What are we aiming for here? What's the purpose of `ItemSeason#month`? Why there are changes unrelated to this method?

vasconsaurus:
I might have misunderstood quite a few things. 😬  I thought we would need a method to associate the month_indexes with the months names, so it would be easier for us to access the names. so that was what I was trying to achieve. how do you think we should go about it?

The other changes were done before, I ended up checking out from one branch to another, so it carried the commits. What would be a best practice?

app/models/item_season.rb
vasconsaurus: why use endless method here?

tcione:
> I might have misunderstood quite a few things. 😬 I thought we would need a method to associate the month_indexes with the months names, so it would be easier for us to access the names. so that was what I was trying to achieve. how do you think we should go about it?

But then why do we need to access the names? As far as I know we only need them to show to the user. Filtering can be done by using the month's index.

However, I do think my comment was a bit obtuse. I was hinting towards having more complete PR titles and descriptions. I've written about it in more detail here: https://github.com/vasconsaurus/produce_in_season/pull/23#pullrequestreview-1000220828

> 
> The other changes were done before, I ended up checking out from one branch to another, so it carried the commits. What would be a best practice?

As we've talked in WhatsApp, the better approach would be always starting from `main`. If something depends on work that lives in a branch, we should consider other strategies, like switching the base branch (which can be done using the "edit" button, on the right side of the PR's title).

vasconsaurus:
> But then why do we need to access the names? As far as I know we only need them to show to the user. Filtering can be done by using the month's index.

I thought this might help with that, but I think I might have overcomplicated things. Although I don't thing we will need it, I updated to use I18n.

As to the PR titles and descriptions, I'll be looking at your suggestions :]

PR #10, opened 2022-06-04
[CHORE] Update Ruby to 3.0.4

tcione:
Ruby 3.0.4 contains 2 important security patches: https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-0-4-released/

Since it's a minor version bump, we should update it.


tcione:
@vasconsaurus I've already merged this one, since it'd block me doing the others. Hope that is fine with you :x

Issue #9, opened 2022-06-03
Create the csv for the seed


vasconsaurus:

vasconsaurus:
@tcione I finished creating the CSV, I'm going to try to create the seed from it.

PR #23, opened 2022-06-08
Seed

vasconsaurus:
I tried to write the seed, but I got a bit confused generating it while making an association between ProduceItem and ItemSeason. It seems to have worked, but I'm not quite sure. What do you think?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -68,11 +68,11 @@ group :development do
 
   # Speed up commands on slow machines / big apps [https://github.com/rails/spring]
   # gem 'spring'
-  gem 'strong_migrations', '~> 1.0.0'
   gem 'brakeman', '~> 5.2.3', require: false
   gem 'rubocop', '~> 1.30', require: false
   gem 'rubocop-rails', '~> 2.14.2', require: false
   gem 'rubocop-rspec', '~> 2.11.1', require: false
+  gem 'strong_migrations', '~> 1.0.0'
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -73,7 +73,6 @@ group :development do
   gem 'rubocop', '~> 1.30', require: false
   gem 'rubocop-rails', '~> 2.14.2', require: false
   gem 'rubocop-rspec', '~> 2.11.1', require: false
-  gem 'strong_migrations', '~> 1.0.0'
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -79,6 +79,7 @@ end
 group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem 'capybara'
+  gem 'database_cleaner-active_record'
   gem 'selenium-webdriver'
   gem 'simplecov', require: false
   gem 'webdrivers'
--- a/app/models/produce_item.rb
+++ b/app/models/produce_item.rb
@@ -6,7 +6,6 @@ class ProduceItem < ApplicationRecord
 
   has_many :item_seasons
 
-  validates :name, presence: true,
-                   uniqueness: true
+  validates :name, presence: true
   validates :category, presence: true
 end
--- a/app/models/produce_item.rb
+++ b/app/models/produce_item.rb
@@ -6,6 +6,6 @@ class ProduceItem < ApplicationRecord
 
   has_many :item_seasons
 
-  validates :name, presence: true
+  validates :name, presence: true, uniqueness: true
   validates :category, presence: true
 end
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,5 +1,7 @@
+# frozen_string_literal: true
+
 # Mark existing migrations as safe
-StrongMigrations.start_after = 20220604103854
+StrongMigrations.start_after = 20_220_604_103_854
 
 # Set timeouts for migrations
 # If you use PgBouncer in transaction mode, delete these lines and set timeouts on the database user
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Mark existing migrations as safe
-StrongMigrations.start_after = 20_220_604_103_854
+StrongMigrations.start_after = '20220604103854'
 
 # Set timeouts for migrations
 # If you use PgBouncer in transaction mode, delete these lines and set timeouts on the database user
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # Mark existing migrations as safe
-StrongMigrations.start_after = '20220604103854'
+StrongMigrations.start_after = 20220604103854 # rubocop:disable Style/NumericLiterals
 
 # Set timeouts for migrations
 # If you use PgBouncer in transaction mode, delete these lines and set timeouts on the database user
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -15,7 +15,7 @@ ItemSeason.destroy_all
 id_arr = []
 
 CSV.foreach(Rails.root.join('lib/produce.csv'), headers: true) do |row|
-  ProduceItem.create({
+  produce_item = ProduceItem.create({
     name: row[0],
     category: row[1]
   })
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -8,11 +8,11 @@
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
 require 'csv'
+require 'database_cleaner/active_record'
 
-ProduceItem.destroy_all
-ItemSeason.destroy_all
+DatabaseCleaner.strategy = :truncation
 
-id_arr = []
+DatabaseCleaner.clean
 
 CSV.foreach(Rails.root.join('lib/produce.csv'), headers: true) do |row|
   ItemSeason.create(
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -15,9 +15,10 @@ DatabaseCleaner.strategy = :truncation
 DatabaseCleaner.clean
 
 CSV.foreach(Rails.root.join('lib/produce.csv'), headers: true) do |row|
-  ItemSeason.create(
+  item = ProduceItem.create!(name: row[0], category: row[1])
+  ItemSeason.create!(
     country_code: row[2],
     month_index: row[3],
-    produce_item: ProduceItem.create(name: row[0], category: row[1])
+    produce_item: item
   )
 end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -6,3 +6,16 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+require 'csv'
+
+CSV.foreach(Rails.root.join('lib/produce.csv'), headers: true) do |row|
+  ProduceItem.create({
+    name: row[0],
+    category: row[1]
+  })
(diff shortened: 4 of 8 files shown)


Gemfile
```diff
@@ -73,6 +73,7 @@ group :development do
   gem 'rubocop', '~> 1.30', require: false
   gem 'rubocop-rails', '~> 2.14.2', require: false
   gem 'rubocop-rspec', '~> 2.11.1', require: false
+  gem 'strong_migrations', '~> 1.0.0'
```
tcione, line 76: This here looks a bit weird. Since we have `strong_migrations` listed as in the dev/test joint group, we should remove it from here.

config/initializers/strong_migrations.rb
```diff
@@ -1,5 +1,7 @@
+# frozen_string_literal: true
+
 # Mark existing migrations as safe
-StrongMigrations.start_after = 20220604103854
+StrongMigrations.start_after = 20_220_604_103_854
```
tcione, line 4: This separation makes no sense (although I do know this is Rubocop's suggestion). We can just ignore this rule here:

```rb
StrongMigrations.start_after = 20220604103854 # rubocop:disable Style/NumericLiterals
```

Or if you'd prefer a sensible format:

```rb
StrongMigrations.start_after = 2022_06_04_10_38_54 # rubocop:disable Style/NumericLiterals
```

Or even not ignoring the rule and transforming it into a string

```rb
StrongMigrations.start_after = '20220604103854'
```

which IMO looks a bit better, since migration versions are not really integers, but dates.

Let me know what you think :)

db/seeds.rb
```diff
@@ -1,8 +1,30 @@
...
 # The data can then be loaded with the bin/rails db:seed command (or created alongside the database with db:setup).
 #
 # Examples:
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+require 'csv'
+
+ProduceItem.destroy_all
+ItemSeason.destroy_all
```
tcione, line 13: This is fine for now, but we could use [database cleaner instead](https://github.com/DatabaseCleaner/database_cleaner). Adding the gem would also help us in testing, since it provides helpful ways for us to clean up database before and after running them.

db/seeds.rb
```diff
@@ -1,8 +1,30 @@
...
+    name: row[0],
+    category: row[1]
+  })
+
+  id_arr << produce_item.id
+
+  ItemSeason.create({
+    country_code: row[2],
+    month_index: row[3],
+    produce_item_id: id_arr.shift
```
tcione, line 28: Why are we adding the ID to an array and then removing it to add it here? I tend to think that using the id directly would be easier (i.e `produce_item.id`). WDYT?

On a sidenote, we could probably simplify this by creating the item directly there:

```rb
ItemSeason.create(
  country_code: row[2],
  month_index: row[3],
  produce_item: ProduceItem.create(name: row[0], category: row[1])
)
```

db/seeds.rb
```diff
@@ -1,8 +1,30 @@
...
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+require 'csv'
+
+ProduceItem.destroy_all
+ItemSeason.destroy_all
+
+id_arr = []
+
+CSV.foreach(Rails.root.join('lib/produce.csv'), headers: true) do |row|
```
tcione, line 17: Up to taste, since it's easy to understand what's going on here, but I'd generally prefer to name the rows. E.g.:
```rb
produce_name, produce_category, country_code, month_index = row
```

Again, up to you opting for it or not :)

lib/produce.csv
```diff
@@ -0,0 +1,535 @@
...
+gobo,green,BR,12
+hortelã,green,BR,12
+manjericão,green,BR,12
+mostarda,green,BR,12
+nabo,green,BR,12
+palmito,green,BR,12
+rabanete,green,BR,12
+rúcula,green,BR,12
+salsa,green,BR,12
+salsão,green,BR,12
```
tcione, line 535: We need to add a new line here :) (more on that [here](https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log/5813359#5813359))

lib/produce.csv
```diff
@@ -0,0 +1,535 @@
+name,category,country_code,month_index
+abacaxi havaí,fruit,BR,1
```
tcione, line 2: More as a thing for the future, but we should considering having these names in a slug format (and maybe in English, not sure how feasible that is).

So instead of `abacaxi havaí`, we'd have `abacaxi_havai`. And then we could add it to our i18n files.

tcione (commented):
Functionally it's looking great :D. I do, however, have a few comments around style and practices.

I'd also like to take this opportunity to ask for more meaningful PR titles and descriptions. Not for me particularly (since I'm aware of the context in which these tasks exist), but as practice for you as a developer. Some articles/videos that I find useful:

- https://artsy.github.io/blog/2020/08/11/improve-pull-requests-by-including-valuable-context/
- https://hugooodias.medium.com/the-anatomy-of-a-perfect-pull-request-567382bb6067
- https://www.youtube.com/watch?v=BbIILUSmSk4

Please let me know if you think this is too much rn :)

tcione:
Unrelated to the PR itself. If you add "closes #9" or "fixes #9" to the PR description, github will automatically close the issue once the PR is merged.

db/seeds.rb
vasconsaurus: I like your solution waaay more. I was trying to figure out how to get the ProduceItem id into ItemSeason and came up with that, I thought it looked a bit wonky but couldn't figure out how to refactor it.

config/initializers/strong_migrations.rb
vasconsaurus: I think turning into a string looks the best

vasconsaurus: Buuut turning it into a string created an error when running migrations, so I'm going to use the first option

lib/produce.csv
vasconsaurus: Adding the slugs is easy, however the names in english I'm not so sure, as there are some local produce as well like cupuaçu, and the different types of mangos or pineapples etc Eg, would we translate abacaxi_havai to pineapple_havai?

I thought about dropping the types, but they are in season in different moments, and in some cases, the types do make a difference, like tangerina murcot vs tangerina cravo

db/seeds.rb
vasconsaurus: I'm trying to run it, but I'm getting an error, not sure if it's related:
`rails aborted!
ActiveRecord::NotNullViolation: PG::NotNullViolation: ERROR:  null value in column "produce_item_id" of relation "item_seasons" violates not-null constraint
DETAIL:  Failing row contains (34, 2, BR, null, 2022-06-09 19:42:24.530141, 2022-06-09 19:42:24.530141).
/Users/manu/code/produce_in_season/db/seeds.rb:18:in `block in <main>'
/Users/manu/code/produce_in_season/db/seeds.rb:17:in `<main>'

Caused by:
PG::NotNullViolation: ERROR:  null value in column "produce_item_id" of relation "item_seasons" violates not-null constraint
DETAIL:  Failing row contains (34, 2, BR, null, 2022-06-09 19:42:24.530141, 2022-06-09 19:42:24.530141).
/Users/manu/code/produce_in_season/db/seeds.rb:18:in `block in <main>'
/Users/manu/code/produce_in_season/db/seeds.rb:17:in `<main>'
Tasks: TOP => db:seed
(See full trace by running task with --trace)`

db/seeds.rb
vasconsaurus: from this, we will get produce_name = row[0] and so on?

db/seeds.rb
tcione: We could create the item before and then assign it.

```rb
item = ProduceItem.create!(name: row[0], category: row[1])
ItemSeason.create!(
  country_code: row[2],
  month_index: row[3],
  produce_item: item
)
```

Added `!` so the program crashes in case there is a validation error.

lib/produce.csv
tcione: No idea about translations, also that's not something we should worry about now, it looks a bit too early. Let's keep as is :)

db/seeds.rb
tcione: Where, though?

vasconsaurus: sorry, forgot to push the latest version

tcione:
@vasconsaurus I think this is the wrong choice here. The problem was not the name being unique, it was creating a produce item in every loop iteration. In the end, we still want every `ProduceItem` to be unique, I think. So the better approach would be:

```rb
  item = ProduceItem.find_by(name: row[0]) || ProduceItem.create!(name: row[0], category: row[1])
  ItemSeason.create!(
    country_code: row[2],
    month_index: row[3],
    produce_item_id: item.id
  )
```

which could be abbreviated to

```rb
  item = ProduceItem.find_or_create_by!(name: row[0], category: row[1])
  ItemSeason.create!(
    country_code: row[2],
    month_index: row[3],
    produce_item_id: item.id
  )
```

Gemfile
tcione: We should remove this line :)

vasconsaurus: sorry, I thought I removed it 😬

db/seeds.rb
vasconsaurus: I tried it, but got a validation error `ArgumentError: '["category", "fruit"]' is not a valid category`

vasconsaurus:
I made some of the small fixes that were missing:
• gemfile: remove strong migrations from the development group
• produce.csv: add line break at the end
• produce.csv: change whitespaces to slugs
• produce_item : added uniqueness back
• seeds: rewrote as suggested

app/models/produce_item.rb
```diff
@@ -6,7 +6,6 @@ class ProduceItem < ApplicationRecord
 
   has_many :item_seasons
 
-  validates :name, presence: true,
-                   uniqueness: true
+  validates :name, presence: true, uniqueness: true
```
tcione, line 9: This a comment that should be handled in a different PR. However it is a good opportunity to talk about uniqueness validation in a Rails app.

`uniqueness` is one of the Rails built-in validations that we should avoid. We can replace it for unique indices instead.

The reasoning behind this is that `uniqueness` does 1 request to the database to check if a record already exists (something like `SELECT 1 FROM produce_items WHERE name = 'an specific name'`). Unique indices, instead, are handled internally by the database, which makes things way faster. In small scale application this does not make much difference, but also there is no reason why we should not apply it.

Now, how do we even do that?

1. We remove the uniqueness validation
2. We add a migration to include unique indices: 
```rb
add_index :produce_items, :name, unique: true
```

We should that soon, because once there are duplicates in the database, things get way harder.

db/seeds.rb
tcione: Ah, I thought it produced an array per iteration. Looking at it I discovered that not. It iteratares through an array of objects (i.e. `#<CSV::Row "name":"abacaxi_havaí" "category":"fruit" "country_code":"BR" "month_index":"1">`).

When we try destructuring, it returns a tuple with header and value (like `["category", "fruit"]`).

An interesting consequence, is that this implies that calling `row.to_h` will give you a hash with the proper keys:
```rb
{"name"=>"abacaxi_havaí", "category"=>"fruit", "country_code"=>"BR", "month_index"=>"1"}
```

So we could use this hash or keep as is. I don't have a preference here.

tcione (approved):
LGTM :D. Thanks for going through the comments!

app/models/produce_item.rb
vasconsaurus: so, for now we merge, and we do a new branch/pr with this migration?

db/seeds.rb
vasconsaurus: this one confused me a bit, is there anything I could read that might help?

PR #27, opened 2022-06-14
feat: activate rails, performance and rspec linters

tcione:
While working on the project I've noticed that I had forgotten to:

a. add `rubocop-performance` to automatically identify code that we write that might not be performant
b. initialize the added rubocop extensions

Thus I set out to active them. Once activated, a new problem surfaced. There were a few offenses caused by these inclusions. So to take this out of our way, I've decided to go on and already fix the offenses.


tcione:
@vasconsaurus I've tried to separate the steps I've went through using commits. Please let me know if you have questions.

spec/models/item_season_spec.rb
```diff
@@ -3,14 +3,15 @@
...
 
+  let!(:produce_item) { ProduceItem.create(name: 'Banana', category: 'fruit') }
+
   it 'is valid with valid attributes' do
     expect(subject).to be_valid
   end
 
   describe 'Associations' do
-    it { should belong_to(:produce_item).without_validating_presence }
+    it { is_expected.to belong_to(:produce_item).without_validating_presence }
```
vasconsaurus, line 17: read today I should use is_expected.to instead of should 😬

vasconsaurus (approved):

PR #30, opened 2022-06-17
Feature – Find an produce item season by produce name

### LINKED ISSUE #28: feature – search produce by name
```
As a user
I want to search produce by name
So I can find the season for my favorite fruit
```

vasconsaurus:
closes #28 

This is the first version of this feature, the user can view all the produce in a table in produce_item#index, from there the user can choose an item, and see it's season's months in produce_item#show.

**Actions and tests**
For this to work, I created produce_items and item_seasons new and show actions. I also implemented a few tests: requests tests for both, and view tests for produce_items (as we currently don't have any item_season views)

**Methods**
1. model method: I made a small model method to take the slugs from names and change them to whitespaces. Since the produce names appear on other views, it might be easier to call a method than to write the logic every time, even if it's tiny.
2. helper method: I thought we could use a helper to format months names, since that will probably appear in other places.

**Rubocop**
I'm having a small problem with a rubocop request, I don't understand what it 's asking?

Screen Shot 2022-06-17 at 13 18 20


spec/views/produce_items/show.html.erb_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    render
+    expect(rendered).to match(/banana/)
+  end
+
+  it 'displays months' do
+    @produce_item = ProduceItem.create!(name: 'banana', category: 'fruit')
+    assign(:item_seasons, [
+      ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 1, country_code: 'br', id: 1),
+      ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 2, country_code: 'br', id: 2)
+    ])
```
tcione, line 17: About the rubocop alert. It expects your array to either be

1
```rb
assign(:item_seasons, [
                         ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 1, country_code: 'br', id: 1),
                         ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 2, country_code: 'br', id: 2)
                       ]
)
```

or 

2
```rb
assign(
  :item_seasons, 
  [
    ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 1, country_code: 'br', id: 1),
    ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 2, country_code: 'br', id: 2)
  ]
)
```

I personally prefer 2, but that's up to you

spec/views/item_seasons/index.html.erb_spec.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
tcione, line 1: If we have no tests yet, we should delete the file

app/helpers/item_seasons_helper.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
```
tcione, line 1: We can delete this file if we don't have plans to use it

app/models/produce_item.rb
```diff
@@ -8,4 +8,8 @@ class ProduceItem < ApplicationRecord
 
   validates :name, presence: true
   validates :category, presence: true
+
+  def name_with_whitespace
```
tcione, line 12: We'll use this to show the produce name to users, correct? If that's the case, we should use a more business-like method name, like `readable_name`. WDYT?

app/models/produce_item.rb
```diff
@@ -8,4 +8,8 @@ class ProduceItem < ApplicationRecord
...
+    name.gsub(/_/, ' ')
```
tcione, line 13: I think you do not need to use a regular expression here. But please check if that's the case.

app/helpers/month_names_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module MonthNamesHelper
+  def format_month_name(month_index)
```
tcione, line 4: In Ruby we generally refer to what is returned by the method, not about what the method is doing. So a more "ruby"-like name would be "formatted_month_name". Of course, this is subjective. So what do you think about the idea?

app/helpers/month_names_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module MonthNamesHelper
```
tcione, line 3: Hm, not sure if I agree with the pattern here. Helpers should be attached to controllers, IMO. I'd like your opinion on this, though.

app/views/item_seasons/index.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>ItemSeasons#index</h1>
```
tcione, line 1: If feels like we do not need the item_seasons views and controller. Any particular reason for us to have them already?

app/views/produce_items/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<p id="notice"><%= notice %></p>
+
+<h1>Produce</h1>
+
+<table>
+    <% @produce_items.each do |produce|  %>
```
tcione, line 6:
```suggestion
    <% @produce_items.find_each do |produce|  %>
```

`find_each` serves the same purpose as `each`, but does not eagerly load all the records. Since this is an easy optimization fix, I think we should apply it.

More on find_each [here](https://apidock.com/rails/ActiveRecord/Batches/ClassMethods/find_each)

app/views/produce_items/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<p id="notice"><%= notice %></p>
+
+<h1> <%= @produce_item.name_with_whitespace %> </h1>
```
tcione, line 3: Looking at this here, it feels a lot like we should store this name in a i18n file (like `pt-BR.yml`) and then use the produce_items names as keys. Does that make sense?

app/views/produce_items/show.html.erb
```diff
@@ -0,0 +1,10 @@
...
+<h3> months </h3>
+
+<% @produce_item.item_seasons.each do |season|  %>
```
tcione, line 6:
```suggestion
<% @produce_item.item_seasons.find_each do |season|  %>
```

spec/views/produce_items/index.html.erb_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'produce_items/index', type: :view do
```
tcione, line 5: I struggle a bit to find utility in view tests. I think they are fine for training but they don't add a lot of value. I'd say that adding e2e tests in the future would be a better bet. WDYT?

tcione (commented):
Nice to see things taking shape :D. I've left a few comments for your consideration.

spec/views/produce_items/show.html.erb_spec.rb
vasconsaurus: thanks! I kept reading and trying to rewrite it, but couldn't figure out what it was asking 😬 
I went with the second one 😄

app/models/produce_item.rb
vasconsaurus: we can use `name.gsub('_', ' ')`, is there an advantage?

app/helpers/month_names_helper.rb
vasconsaurus: I had my doubts regarding the best place for this helper, as I haven't worked with them much. What would you suggest?

app/views/item_seasons/index.html.erb
vasconsaurus: nope, I thought we were going to need them, but that is not the case

spec/views/produce_items/index.html.erb_spec.rb
vasconsaurus: What kind of tests do you find usefull? Do you think we shouldn't have any view tests at all? Regarding models and controllers tests, would you remove/add something? Ho would you go abtou writing e2e tests?

app/views/produce_items/index.html.erb
vasconsaurus: Now my view test is failling because I can't use the find_each on an array. I'm trying to figure out how to rewrite it, but so far no luck. How could I rewrite it, should we keep this test anyway?

app/models/produce_item.rb
tcione: Simple string replacement has better performance than regular expressions :D

app/views/produce_items/index.html.erb
tcione: You should assign an ActiveRecord relation instead of an array, since this is what will happen in the real application. So you'd do something like:

```rb
ProduceItem.create!(...)
ProduceItem.create!(...)
assign(:produce_items, ProduceItem.all)
```

spec/views/produce_items/index.html.erb_spec.rb
tcione: Tests are a topic in which opinions vary a lot. I'll start on what (almost) everyone agrees and then highlight the places in which I have a different perspective.

So we all usually use what is called the test pyramid:



The quantity of tests you should have is equal (in proportion) to the width of the step. Thus we should have mostly unit tests and few e2e tests.

Now what is a unit tests is a controversial topic, the classical definition states that these tests are ISOLATED, meaning there is no interaction with anything but the piece of code you are testing. This is not applicable to Rails, since we do not mock the database to make our model tests (although we still call them "unit tests"). So from my perspective, IN A RAILS application, I use the following definitions:

- Unit tests: tests that test a single class. Database interactions are fine as long we're not testing two different objects
- Integration tests: tests that test a combination of classes or contain side effects. Testing controllers is a good example, but there are also others, like testing a [service object](https://www.toptal.com/ruby-on-rails/rails-service-objects-tutorial) that joins several other service objects
- e2e tests: tests that actually boot a browser, click on stuff and test behavior (`capybara` does those, `cypress` does them as well)

Now there is the question of what should be tested:

- Everything inside models must be tested
- Every service object must be tested
- Controller tests are painful to maintain (and also made useless by e2e tests). We should only make these to test API endpoints
- View tests are useless if we have e2e tests. Testing what is rendered has little value IMO
- e2e tests are very important, but should be created with reservation, since they are very very slow. However we should test all our important flows (meaning user stories/scenarios).

Hope that sheds some light into this topic :x

OBS.: Read [this article](https://martinfowler.com/articles/practical-test-pyramid.html) to get a more traditional perspective

app/helpers/month_names_helper.rb
tcione: Putting the helper method under the helper attached to a specific controller (in this case ProduceItemHelper)

app/views/produce_items/show.html.erb
vasconsaurus: [is this what you meant?](https://github.com/vasconsaurus/produce_in_season/blob/26735d66e0defcb526958cf28b3f86c9063be9cb/config/locales/pt-BR.yml)

if so, why should we do this?

vasconsaurus:
> Nice to see things taking shape :D. I've left a few comments for your consideration.

Alt Text

app/views/produce_items/show.html.erb
tcione: > [is this what you meant?](https://github.com/vasconsaurus/produce_in_season/blob/26735d66e0defcb526958cf28b3f86c9063be9cb/config/locales/pt-BR.yml)

Yes

> if so, why should we do this?

Replacing `_` assumes that that's the only thing we did to [slugify](https://www.semrush.com/blog/what-is-a-url-slug/), while we do know that these version should not contain characters that are not included in `[a-zA-Z0-9-_]` (if there are those, we should either replace or just use the actual name in the database).

Having a dictionary avoids these pitfalls. No logic is needed, we just map keys to values. There is also a nice bonus that it's easy to translate if that's ever the case.

app/helpers/month_names_helper.rb
vasconsaurus: I moved it to ProduceItemHelper, and deleted the MonthNamesHelper

app/views/produce_items/show.html.erb
vasconsaurus: huuuum, would we do something like this? 

```
    <% @produce_items.find_each do |produce|  %>
      <tr>
        <td><%= I18n.backend.translations[:'pt-BR'].key produce.name %></td>
...
```

Instead of using readable_name? Though not sure if this is what you thought, I tested it here, but it doesn't seem to work on names that have more than one word. (the first colum in the image)

Screen Shot 2022-06-17 at 18 29 37

tcione: > huuuum, would we do something like this?

Yeap. I'd actually write like `I18n.t(produce.name)`. Even better would be to prefix it, like:

```yml
pt-BR:
  produce_items:
    names:
      abacaxi_havai: 'Abacaxi Havaí',
      ...
```

Then the snippet would become `I18n.t("produce_items.names.#{produce.name}")`.

I think it's fine to put that (`I18n.t...`) inside `ProduceItem#readable_name`, that way we have a nice abstraction :).

BTW, putting the language in the string call (`translations[:'pt-BR']`) goes against the i18n lib purpose. We should rely on the value we set in `config/application.rb` using `config.i18n.default_locale` (it is `:en` by default, but we could change it to pt-BR).

>  I tested it here, but it doesn't seem to work on names that have more than one word. (the first colum in the image)

That's due to using special characters in the yaml keys.

vasconsaurus: Since all our produce is in portuguese, and the seasons are local, I changed our default to be pt-BR. I also translated other english words. 

I think it makes more sense to start with the brazilian version and in the future make the german and english  versions. Also, to that, we'll have to find out the local seasons, and there will be produce we will have to introduce and remove. Right now there are a lot of brazilian produce in our list, that wouldn't make sense for a german user.

tcione: > Do you think we should abstract both of these into methods?

Nope, we extract them if/when we need to.

app/views/item_seasons/index.html.erb
tcione: Nope, let's delete everything that we do not use.

Gemfile
```diff
@@ -60,6 +60,8 @@ group :development, :test do
 
   gem 'brakeman', '~> 5.2.3', require: false
   gem 'bullet', '~> 7.0.2'
+  gem 'faker'
+  gem 'factory_bot_rails'
```
tcione, line 64:
```suggestion
  gem 'faker', '~> 2.21.0'
  gem 'factory_bot_rails', '~> 6.2.1'
```

In the long run, using pessimistic versioning (meaning that we specify all [SEMVER](https://semver.org/) slots, will help us in upgrading the system. Whenever we run `bundle update xyz` it will update all dependencies according to the rules we set, which avoids unexpected surprises (like bugs caused by upgrading major versions without being aware).

I do know, though, that it sounds very pedantic, so I'll leave the decision to use it or not to you :)

app/views/produce_items/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<p id="notice"><%= notice %></p>
+
+<h1>Items</h1>
+
+<table>
+    <% @produce_items.find_each do |produce|  %>
+      <tr>
+        <td><%= I18n.t("produce_items.names.#{produce.name}") %></td>
+        <td><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
+        <td><%= link_to 'ver', produce %></td>
```
tcione, line 10: Should we move hardcoded strings to our `pt-BR.yml` file?

config/locales/pt-BR.yml
```diff
@@ -0,0 +1,556 @@
...
+      damasco_estrangeiro: damasco estrangeiro
+      figo: figo
+      framboesa: framboesa
+      graviola: graviola
+      kiwi_estrangeiro: kiwi estrangeiro
+      laranja_pera: laranja pera
+      lichia: lichia
+      limão_taiti: limão taiti
+      maçã_nacional_fuji: maçã nacional fuji
+      maçã_estrangeira_red_delicious: maçã estrangeira red delicious
```
tcione, line 521: Keys should not contain special characters. So let's restrict to `a-z A-Z 0-9 _ -`. Thus:
- `maçã_estrangeira_red_delicious` becomes `maca_estrangeira_red_delicious`
- `abacate_breda/margarida` becomes `abacate_breda_margarida`
- `abóbora_d'água` becomes `abobora_d_agua`
- etc

spec/factories/produce_item.rb
```diff
@@ -0,0 +1,8 @@
+require 'faker'
+
+FactoryBot.define do
+  factory :produce_item do
+    name { Faker::Food.fruits }
```
tcione, line 5: We have a unique constraint on `name`, so this could cause us trouble in the long run. To ensure uniqueness, we must use FactoryBot sequences:

https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences

On a sidenote, I generally prefer using values that I know, so I'd opt for something like `banana` instead of `Faker::Food.fruits`. However, this is a matter of taste, so up to you deciding for Faker or a hardcoded value.

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+# describe ProduceItemsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe ProduceItemsHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
```
tcione, line 16: We should be testing `ProduceItemsHelper#formatted_month_name` here.

spec/models/item_season_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe ItemSeason, type: :model do
-  let!(:produce_item) { ProduceItem.create(name: 'Banana', category: 'fruit') }
+  let!(:produce_item) { create(:produce_item) }
   subject { described_class.new(produce_item_id: produce_item.id, month_index: 1, country_code: 'br', id: 1) }
```
tcione, line 7: Hardcoding the `id` will lead to [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html). Any reason for defining it here?

spec/views/produce_items/show.html.erb_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    render
+    expect(rendered).to match(/banana/)
+  end
+
+  it 'displays months' do
+    @produce_item = ProduceItem.create!(name: 'banana', category: 'fruit')
+    assign(
+      :item_seasons,
+      [
+        ItemSeason.create!(produce_item_id: @produce_item.id, month_index: 1, country_code: 'br', id: 1),
```
tcione, line 17: We shouldn't hardcode `id`s. They lead us to [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html).

Gemfile
vasconsaurus: I don't think it sounds pedantic at all! Seems like a good practice to me :]

app/views/produce_items/index.html.erb
vasconsaurus: you mean like 'show', 'back' etc right? makes sense

spec/models/item_season_spec.rb
vasconsaurus: No reason, it was a mistake 😬

spec/factories/produce_item.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'faker'
+
+FactoryBot.define do
+  factory :produce_item do
+    sequence(:name) { Faker::Food.fruits }
```
tcione, line 7: I feel like this is still susceptible for collisions. You probably should do something like:
```rb
sequence(:name) { |n| "#{Faker::Food.fruits}-#{n}" }
```

vasconsaurus: edited as you suggested, but I'm going to read a bit on sequences in FactoryBot :]

vasconsaurus: so sequence adds n(number) to our fruit name, that way each is unique right? Somehow I misunderstood it a bit at first.

spec/features/produce_items_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  it 'expects link to be visible' do
+    visit(produce_items_path)
+    find_link('ver', visible: :all).visible?
+  end
+
+  it 'expects link to navigate to show and back to index' do
+    visit(produce_items_path)
+    click_link('ver')
```
tcione, line 26: Hm, relying on text matching make the tests a bit too fragile (if we change content, tests will break). We should use css and match by classes of ids. WDYT?

vasconsaurus: Agreed! Will work on this tomorrow :]

tcione (approved):
Ship it

Thanks for you unyielding patience w/ all the comments :D

app/helpers/item_seasons_helper.rb
```diff
@@ -0,0 +1,4 @@
...
+
+module ItemSeasonsHelper
```
tcione, line 3: @vasconsaurus Before merging, please do a final check to delete files that are empty/we're not using :)

PR #32, opened 2022-06-27
feat: choose a month and view it's produce

### LINKED ISSUE #29: feature – search by month
```
As a user
I want to search by month
So I know which produce can I buy in that time of year
```

vasconsaurus:
closes #29 

I tried to start by writing tests, but the logic I thought at first didn't work for the feature. I wrote as if we were creating seasons in the index, but we are only pulling our 12 months from our dictionary and redirecting to the month with it's produce.

As I moved foward and wrote more tests, I realized something was off, as they kept failling, and I couldn't write them properly. The problem was I was trying to pass month_index as a param, but using id. So I changed our item_seasons to use the month_index as param. I also added a helper to get the month index from the month name.

That worked, but I wrote: 
`    @item_season = ItemSeason.find_by_month_index(params[:month_index])`
Which the rubocop keeps yelling about 😬 , when I change to `find_by`, as it suggests, it breaks

### **not directly realted to this feature**

- I deleted the tests folder, as we are using rspec
- Added links to the months in the produce_items show (as we have the item seasons now)

### **related to testing**
I'm still not too sure about the tests. Could you point out which ones you think are usefull, which aren't, and some ways I could make them better?


app/views/item_seasons/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<p id="notice"><%= notice %></p>
+
+<h1 class='page__title'> <%= I18n.t('item_season.index.title') %> </h1>
+
+<table class='produce-table'>
+  <% I18n.t('date.month_names').each do |month_name|  %>
```
tcione, line 6: Hm, I feel like logic here is somehow inverted. We are getting an array of months from translations to then translate back to an index. This also forces you to handle the fact that `month_names` has `nil` as the first item.

I'd prefer to get the indices from a model to then derive a name from it. For example:
```erb
<% ItemSeason::MONTHS.each do |month_index| %>
  <%= I18n.t('date.month_names')[month_index] %>
<%  end %>
```

WDYT?

app/views/item_seasons/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<p id="notice"><%= notice %></p>
+
+<h1 class='page__title'> <%= formatted_month_name(@item_season.month_index).capitalize %> </h1>
+
+<table class='produce-table'>
+  <% ItemSeason.includes([:produce_item]).where(month_index: @item_season.month_index).find_each do |season| %>
```
tcione, line 6: `ItemSeason.includes([:produce_item]).where(month_index: @item_season.month_index)` should probably be assigned in the controller

app/views/produce_items/index.html.erb
```diff
@@ -5,7 +5,7 @@
 <table class='produce-table'>
     <% @produce_items.find_each do |produce|  %>
```
tcione, line 6: 🍬

config/locales/pt-BR.yml
```diff
@@ -2,11 +2,16 @@ pt-BR:
   action:
     back: voltar
     show: ver
+    back_to_produce_list: voltar a lista de items
+    back_to_months_list: voltar a lista de meses
```
tcione, line 6:
```suggestion
    back_to_produce_list: voltar à lista de items
    back_to_months_list: voltar à lista de meses
```

spec/features/produce_items_spec.rb
```diff
@@ -30,7 +30,7 @@
   it 'expects link to navigate to show and back to index' do
     visit(produce_items_path)
     find('table tr td a.link_action').click
-    find('.link_action').click
+    all('.link_action').last.click
```
tcione, line 33: Should we have a specific class or id for this link? Relying on fetching the last of its class sounds fragile to me.

spec/features/season_items_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'ProduceItems', type: :feature do
+  let!(:produce_item) { ProduceItem.create(name: 'carambola', category: 'fruit') }
```
tcione, line 6: Since we use it in the `before` block, there's no need to use `let!`. `before` will make sure we have this item by the time we create `ItemSeason`

spec/features/season_items_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  it 'expects index to have a month' do
+    visit(item_seasons_path)
+    expect(page).to have_css('table tr td.produce-table__month')
+  end
+
+  it 'expects the page to have 12 links' do
+    visit(item_seasons_path)
+    links = page.all(class: ['link_action']).length
+    expect(links).to match(12)
+  end
```
tcione, line 24: `12 links` maps to what exatcly? Would it be better to translate to business logic? (i.e.: has one link per month)

spec/features/season_items_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  it 'expects link to navigate to show' do
```
tcione, line 26: Which link?

spec/features/season_items_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    visit(item_seasons_path)
+    first('table tr td a.link_action').click
+    expect(page).to have_css('table tr td.produce-table__name')
+  end
+
+  it 'expects link to navigate to show and back to index' do
```
tcione, line 32: Which link? The description seems a bit confusing to me.

tcione (commented):
I tend to think controller and view tests are not useful. I'd generally cover these pieces using system tests (here named as `features`). That said, I propose we continue writing them for practice :)

app/controllers/item_seasons_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  def index
+    @item_seasons = ItemSeason.all
+  end
+
+  def show; end
+
+  private
+
+  def set_item_season
+    @item_season = ItemSeason.find_by_month_index(params[:month_index])
```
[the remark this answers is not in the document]
vasconsaurus: so, I looked around a bit and from what I understand: metaprogramming is writing programs that write programs. The first one worked because rails generated a bunch of stuff under the hood, but it has a worse perfomance. I read also that it might not be great for readability, searchability and security?

app/views/item_seasons/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<p id="notice"><%= notice %></p>
```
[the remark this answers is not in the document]
vasconsaurus: should we use `flash[:notice]`?

app/views/item_seasons/index.html.erb
vasconsaurus: I completely agree, I had doubts myself, and thought it was a rather roundabout way to go at it. This makes much more sense.

The only thing I'm suggesting is using our `formatted_month_name` helper instead of `I18n.t('date.month_names')[month_index]`. I think it's more readable, and it's already available to us :]

spec/features/produce_items_spec.rb
vasconsaurus: Agreed. Added a back class to all back links

spec/features/season_items_spec.rb
vasconsaurus: But in the `let!` we are creating a `produce_item`, and in the `before` we create an `item_season` using that `produce_item.id`. If I remove it the tests break.

app/views/item_seasons/index.html.erb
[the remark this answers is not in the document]
tcione: not necessarily. I think `notice` will still be there. However, if you want to make it super flexible (like considering other types of flash messages), then yes, let's go for `flash[]`

app/views/item_seasons/index.html.erb
tcione: Makes sense. Then we should move the helper to `application_helpers`.

app/views/item_seasons/show.html.erb
tcione: I don't think the name communicates what's stored in the instance variable. In the end, you are still referring to a sequence of item seasons. The fact that you later find the associated produce shouldn't matter in the naming. Does that make sense?

tcione:
> @tcione
> 
> This change was just a result of me messing around trying to figure this out. Since I changed the logic to the controller and changed the instance variable name both #show tests broke. I'm getting this error:
> 
> ```
>      Failure/Error: <% @produce_item.find_each do |produce| %>
>      
>      ActionView::Template::Error:
>        undefined method `find_each' for nil:NilClass
> ```
> 
> So I'm guessing my problem is with how I'm creating the variables. I'll keep poking around. But could you suggest maybe what would be a good way to debug rspec?

This is the a problem similar to the one you faced in a previous PR. [You are assigning an array](https://github.com/vasconsaurus/produce_in_season/blob/59710023f505c66819f9162f6cb4ac600d954ed8/spec/views/produce_items/show.html.erb_spec.rb#L20-L23) in a place in which you actually expect an `ActiveRecord` relation. The correct thing would be assigning `ItemSeason.(where|all)` instead.

spec/features/season_items_spec.rb
vasconsaurus: I thought you meant completely removing it. Leaving it as below works:
`RSpec.describe 'ProduceItems', type: :feature do
  let(:produce_item) { ProduceItem.create(name: 'carambola', category: 'fruit') }

  before { ItemSeason.create(produce_item_id: produce_item.id, month_index: 1, country_code: 'br') }`
  
So the advantage would be that let is lazy-evaluated: it is not evaluated until the first time
the method it defines is invoked, while let! is invoked before each example?

app/views/item_seasons/show.html.erb
tcione: I still think you are not fetching produce, but seasons. If you care about produce item, you should invert the query:

`ProduceItem.joins(:item_seasons).where(item_seasons: { month_index: @item_season.month_index })` and then assign to something like `@produce_items` or `@season_produce_items`

I think I misanderstood in the first comment, I should have suggested this from the get-go.

spec/features/season_items_spec.rb
tcione: Yeap, exactly. As a rule of thumb, avoid eager evaluation unless you really need it.

vasconsaurus:
> That said, I'm not sure what do you mean when you mention debugging capybara. You mean seeing the browser at work?

I mean that when I looked at the docs for capybara there were a few debugging tools, like save_and_open_page. That actually helped me quite a bit. With the other parts of rspec I didn't find something like that. I didn't mean to separate them though.

I read a bit about using pry with rspec, I think I'm going to probably try that or byebug

app/views/layouts/application.html.erb
```diff
@@ -11,6 +11,7 @@
   </head>
 
   <body>
+    <% if notice %><aside id="notice"><%= notice %></aside><% end %>
```
tcione, line 14: **suggestion:** changing `aside` to `div`

Not sure if that's the proper call, since there's often space for discussion within HTML semantics, but I see aside as something related to the main content of the page, which is often not the case with `notice` (it could be, for instance, a "login" notification".

Source:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside

spec/features/season_items_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  it 'expects index to have a month' do
+    visit(item_seasons_path)
+    expect(page).to have_css('table tr td.produce-table__month')
+  end
+
+  it 'expects the page to have one link per month' do
+    visit(item_seasons_path)
+    links = page.all(class: ['link_action']).length
+    expect(links).to match(12)
```
tcione, line 23: **suggestion:** We could use `ItemSeason::MONTHS.length` instead of `12`

I think this would carry more meaning.

spec/views/item_seasons/show.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+require 'rails_helper'
+
+RSpec.describe 'item_seasons/show', type: :view do
+  let!(:item_season_produces) { ProduceItem.create(name: 'carambola', category: 'fruit') }
+  let!(:item_season) { ItemSeason.create(produce_item_id: item_season_produces.id, month_index: 1, country_code: 'br') }
+
+  it 'display month name' do
+    assign(:item_season, item_season)
+    assign(:item_season_produces, ProduceItem.joins(:item_seasons)
+    .where(item_seasons: { month_index: item_season.month_index }))
```
tcione, line 12: **nitpick:**  Indentation looks off to me

We could move `ProduceItem.joins(:item_seasons).where(item_seasons: { month_index: item_season.month_index })` to a variable. If you prefer not to, I'd usually go for

```rb
    assign(
      :item_season_produces, 
      ProduceItem.joins(:item_seasons).where(item_seasons: { month_index: item_season.month_index })
    )
```

tcione:
More as an FYI, since I'm changing the comment style. We are currently adopting [Conventional Comments](https://conventionalcomments.org/) in my team, so I'm using this project to get used to it as well :p

app/helpers/item_seasons_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module ItemSeasonsHelper
+  def month_name_to_index(month_name)
```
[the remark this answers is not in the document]
vasconsaurus: Nope, we should remove it, I forgot to do this yesterday

spec/views/item_seasons/show.html.erb_spec.rb
vasconsaurus: Agreed, but the way you suggested makes rubocop complain about 'too many lines'. What do you think about this? 
```
    assign(:item_season_produces, ProduceItem.joins(:item_seasons)
                                             .where(item_seasons: { month_index: item_season.month_index }))

```

I tried moving it to a variable, but couldn't get it to work

tcione: TBH I think we should tweak Rubocop's configuration to be less strict in this aspect. Would you mind sharing the error here?

vasconsaurus:
```
Offenses:

spec/views/item_seasons/show.html.erb_spec.rb:17:3: C: RSpec/ExampleLength: Example has too many lines. [7/5]
  it 'displays the produces' do ...
```

We had this problem once, and we excluded the file in the rubocop.yml, but maybe you should change Rubocop's configuration regarding this rule.

```
RSpec/ExampleLength:
  Exclude:
    - 'spec/views/produce_items/show.html.erb_spec.rb'
```

vasconsaurus: do you think we should set a new max? 11 would work for our current long tests. Or maybe should we disable this test?

vasconsaurus: I was messing around a bit, and tried using `CountAsOne`, it takes our `produce_items/show.html.erb_spec.rb` from 11 lines to 8, so doesn't solve it, and doesn't do anything for the other two. Still, it could be helpfull. 

Right now all our problems are in the views, we could exclude that folder. But this problem could appear again in the future in another test.

Right now I have something like this:
```
RSpec/ExampleLength:
  CountAsOne:
    - array
    - hash
    - heredoc
  Max: 5
  Exclude:
    - 'spec/views/**/*.rb' 
```

WDYT?

tcione: I like it :). However `Max: 5` seems very little to me. WDYT about moving it to 10?

vasconsaurus: another solution could be changing the tests, this works:

```
  before do
    assign(:item_season, item_season)
    assign(
      :item_season_produces,
      ProduceItem.joins(:item_seasons).where(item_seasons: { month_index: item_season.month_index })
    )
  end

  it 'display month name' do
    render
    expect(rendered).to match(/Janeiro/)
  end

  it 'displays the produces' do
    render
    expect(rendered).to match(/Carambola/)
  end
```

tcione: Hm, the structure you showed in the comment before looks better to me (since you are DRYing away the assigns). Let's go with it AND w/ the increased limit.

vasconsaurus: I was reading how in tests sometimes repetition was warranted if it promoted readability. I think here the tests are super simple, and separating the assigns doesn't really make it harder to read. But I was wandering which way was best to write it, and ended leaving it there 😬

tcione (approved):
LGTM

PR #34, opened 2022-07-18
feat: main page/root and 404

### LINKED ISSUE #33: feature - main page
```
As a user
I want to see the main page
So I know what this is app about and I can find my way in the app
```

vasconsaurus:
closes #33 

This is the first draft for our main/home page. It should present what the purpose of our app is, and be a gateway to our searches. It is still very simple, and has no styles.

Since I built the controller for the main page, and it's a "pages" controller, I decided to go ahead, and build the structure for our one off pages (such as about, and errors etc). From those, I built the 404.


config/routes.rb
```diff
@@ -1,9 +1,14 @@
 # frozen_string_literal: true
 
 Rails.application.routes.draw do
+  root to: 'pages#index'
```
tcione, line 4: **suggestion**: I generally prefer to have a controller for the home page. `pages#index` implies that we'll see a list of possible pages for `pages#show`. WDYT?

config/locales/pt-BR.yml
```diff
@@ -27,6 +27,11 @@ pt-BR:
     - outubro
     - novembro
     - dezembro
+  produce_item:
+    index:
+      title: Items
```
tcione, line 32:
```suggestion
      title: Itens
```

app/views/pages/index.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1><%= I18n.t('root.title') %></h1>
+<h3><%= I18n.t('root.about') %></h3>
+
+<p><%= link_to I18n.t('action.search_by_month').capitalize, item_seasons_path, class: "link_action" %></p>
```
tcione, line 4:
```suggestion
<p><%= link_to I18n.t('action.search_by_month').capitalize, item_seasons_path, class: 'link_action' %></p>
```

config/routes.rb
vasconsaurus: Agreed! So i generated a new controller for the homepage and used show, going to keep pages for our miscellaneous pages.

config/locales/pt-BR.yml
vasconsaurus: I forgot how to portuguese 😆

vasconsaurus:
@tcione two things:

1. Brakeman
In the pages controller: I had moved the params in the action for a test, forgot to move them back, brakeman complained. I moved them back. But I keep on getting the same warning?

Screen Shot 2022-07-29 at 12 21 09
Screen Shot 2022-07-29 at 12 21 17

2. Pages Controller
So the pages with slugs for error pages and other stuff, it was something we did (I think) for singyeTalkie, and it worked fine, but now I'm not sure if that is the way to go? It only works in a few cases, I'm posting a few images to make it a bit clearer

Screen Shot 2022-07-29 at 11 35 16
Screen Shot 2022-07-29 at 11 35 46
Screen Shot 2022-07-29 at 11 36 15

app/controllers/home_page_controller.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class HomePageController < ApplicationController
+  def show; end
```
tcione, line 4: **idea**: `show` normally suggests an entity in the database. I have a bias towards `index`. WDYT about this topic?

app/controllers/pages_controller.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class PagesController < ApplicationController
+  before_action :set_slug, only: %i[show]
```
tcione, line 4: **suggestion**: we can cut the filter (`only`)

app/controllers/home_page_controller.rb
vasconsaurus: I misunderstood your first comment in regards, and thought I should replace it with show 🤦‍♀️ But, yeah, while I was doing it I thought it was a bit weird to use it this way derp.

tcione:
## Brakeman complaint

You are rendering a file from use input. This is a potential source of danger for it's exploitable. Someone can inject some sort of string with unexpected outcomes.

For instance `http://localhost:3000/../../Gemfile` will deliver a 500 error for trying to call `render 'pages/../../Gemfile`. In the very least, this might reveal the files on your server. In the very worst, this might allow for [remote code execution](https://www.bugcrowd.com/glossary/remote-code-execution-rce/).

To circumvent that, you could generate an allowlist (static or dynamic). So you'd have something like:
```ruby
ALLOWED_SLUGS = %w[about]

def show
  slug = params[:slug]
  return render 'pages/404', status: 404 if !ALLOWED_SLUGS.include?(slug)
  
  render "pages/#{slug}"
end
```

This would not solve the alert, but would solve the security issue it points to. To solve the alert you would have to do something like:
```ruby
# This could be made dynamic by listing files inside `pages/`
PAGES =  {
  'about' => 'pages/about'
}

def show
  page = PAGES[params[:slug]]
  return render 'pages/404', status: 404 if page.blank?
  
  render page
end
```

## Pages Controller

You presented 3 examples:

1. `/404`, which renders Rails' default 404 page. This happens because rails ships with a `404.html` file under `public/` and it takes precedence over whatever routes you have. To get around this, you can follow [this article[(https://www.mintbit.com/blog/custom-404-500-error-pages-in-rails)
2. `/3`, which renders your custom 404. I assume, this is the one that works as you expect, so I won't get into it
3. `/produces_item/e`, you have an active record here, it pops out before the router can make any informed decision. `ProduceItem.find(id)` raises an exception if the record does not exist. To get a silent failure you should use `ProduceItem.find_by(id: id)`. If you'd like not finding a record to always trigger a 404 (which I'd recommend against, since this is the sort of behavior that's better of being bespoke), you can rescue the specific error inside the application controller and render a 404 page, [like in here](https://stackoverflow.com/a/6773038)

Hopefully this made things a bit more clear :D. Let me know if you still have questions.

vasconsaurus:
@tcione thank you for your patience 🙏 
I decided to remove the whole slugs thig for now, honestly I should have only worked in the main page for this one. Let's keep it simple while we are able to.

tcione (changes requested):
Code-wise LGTM. I do have one question:

1. Why are the images already here?
2. We should run these images through an optimizer (like [this one](https://imageoptim.com/mac]))

vasconsaurus:
> Code-wise LGTM. I do have one question:
> 
>     1. Why are the images already here?
> 
>     2. We should run these images through an optimizer (like [this one](https://imageoptim.com/mac%5D))

Because of how I used to organize my work as a graphic designer 🥲 I was doing the image research, and thought I should save in the project images folder. As I used to do. But later I realized I ended up saving the images to git, to this branch, and it slows things down. So I was wandering, what are the best practices when working with images? What do you suggest we do here? Only run through the optimizer? Should we delete them and and ammend the commit?

tcione:
My suggestions around best practices:
- Only add images that are being used by the application to the repo
- Optimize all images for internet usage
- Make sure images are just as big as needed (in other words, as small as your layout allows :p)

tcione (approved):
LGTM

PR #44, opened 2022-08-22
Feature/main page frontend

vasconsaurus:
I finished the first version for the main page's frontend.

Since the asset pipeline is giving us some trouble, I uploaded the images we are using to oneDrive, and linked from there. This way we can at least validate if the way the css is ok, and move on to the other pages, while we work on the asset pipeline.

@tcione tell me if you think the way I organized things is ok/makes sense, if we can continue like this. As always, I'm open to suggestions :]


app/assets/stylesheets/components/_buttons.css
```diff
@@ -0,0 +1,19 @@
+.button-holder {
```
tcione, line 1: **issue**: Naming is a bit inconsistent. `button-holder` vs `yellow-button`. Better would be `button-holder` vs `button-yellow`.

Normally I like to follow a few guidelines when doing css (feel free to challenge them):

1. One file per component (so `_yellow-button.css` and `_button-holder.css`
2. extensibility, I think we'd benefit from having a `button` class with basic rules and derive from that, maybe using BEM principles:
```css
.button {
  display: flex;
  justify-content: center;
  align-items: center;
}

.button--yellow {
    background-color: var(--yellow);
}
```

```html
<a class="button button--yellow"></a>
```

app/assets/stylesheets/components/_links.css
```diff
@@ -0,0 +1,9 @@
+a, a:visited {
```
tcione, line 1: **suggestion:** following up on how I organize CSS, I always have a file called `_base.css` for all "classless" styles. WDYT?

app/assets/stylesheets/components/_navbar.css
```diff
@@ -0,0 +1,52 @@
...
+    gap: 0;
+    align-items: flex-end;
+    text-align: right;
+  }
+}
+
+.nav-right p {
+  font-weight: 900;
+  font-size: 12pt;
+  line-height: 13pt;
```
tcione, line 51: **issue:** avoid using non-screen units. Favor `px`, `em` and `rem`. `rem` is better (https://www.sitepoint.com/understanding-and-using-rem-units-in-css/).

I personally prefer defining a base unit in percentage at the body and using `rem` for everything else. However, feel free to use `px`, since it's simpler.

app/assets/stylesheets/components/_navbar.css
```diff
@@ -0,0 +1,52 @@
...
+  color: var(--blue);
+  text-align: center;
+}
+
+.design-elements {
+  font-family: Eds Market Design Elements;
+}
+
+.nav-right {
+  display: flex;
```
tcione, line 28: **thought**: I feel like we're abusing `flex`. Are there places where we could get away with `block` or any default layout model?

app/assets/stylesheets/config/_colors.css
```diff
@@ -0,0 +1,23 @@
...
+  color: var(--red);
+}
+
+.blue-txt {
+  color: var(--blue);
+}
+
+.white-txt {
+  color: white;
+}
```
tcione, line 23: **suggestion**: I'd prefer to follow `txt-white`. That said, `txt` maybe abstracts too much. WDYT about `color-white`. Moreover, do we need these?

app/assets/stylesheets/config/_spaces.css
```diff
@@ -0,0 +1,64 @@
+/* paddings */
+.p-top-15 {
```
tcione, line 2: **idea:** I tend to prefer using numbers for standardization. So `p-top1`, `p-top2`, etc. That way you think in proportions instead of sizes.

You can see an example in my vars here: https://github.com/tcione/tales.dev/blob/main/sass/_base.scss

tcione (commented):
Nice to see things taking shape. I did a quick review (will do a more detailed one over the weekend), but I think there are already some topics for you to consider.

When it comes to PR description, would you mind putting some screen captures of the outcome? It'd make it easier for me to evaluate without running the code :D

vasconsaurus:
> Nice to see things taking shape. I did a quick review (will do a more detailed one over the weekend), but I think there are already some topics for you to consider.
> 
> When it comes to PR description, would you mind putting some screen captures of the outcome? It'd make it easier for me to evaluate without running the code :D

No problem! I'm sorry, it completely slipped my mind. I made two videos, one with a bigger width, and one on mobile. Let me know if something like this helps.

https://user-images.githubusercontent.com/87862340/187267822-6ddbab84-bfb9-41f5-9102-24433609da32.mov

https://user-images.githubusercontent.com/87862340/187267950-65fbdee8-3aaf-4745-8058-c933300b4452.mov

app/assets/stylesheets/application.css
```diff
@@ -13,3 +13,32 @@
...
+
+@import "components/buttons";
+@import "components/footer";
+@import "components/links";
+@import "components/navbar";
+@import "config/colors";
+@import "config/fonts";
+@import "config/rotations";
+@import "config/spaces";
+@import "pages/homepage.css";
```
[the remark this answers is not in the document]
vasconsaurus: I have not. Would I check this by checking the server? Could you maybe explain to me what I'm looking for?

I tested with sprocket's require but it seems to take longer, could be me misreading it though:
– with require:
`Completed 200 OK in 24ms (Views: 23.4ms | Allocations: 18234)`
– with import:
`Completed 200 OK in 4ms (Views: 3.6ms | Allocations: 4136)`

Also, not sure if it's related, but what do you think about working with sass, adding the 'sassc-rails' gem?

app/assets/stylesheets/application.css
```diff
@@ -13,3 +13,32 @@
...
+
+*, *::before, *::after {
+  box-sizing: border-box;
+  margin: 0px;
+  padding: 0px;
+  /* border: 0.5px solid blue; */
```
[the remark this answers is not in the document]
vasconsaurus: I added it as you suggested, but what are the benefits to it?
(it is for debugging 🙂 )

app/assets/stylesheets/components/_buttons.css
vasconsaurus: agreed. sometimes I still find myself confused regarding how to name things 😬

app/assets/stylesheets/components/_navbar.css
vasconsaurus: In this case I do think it serves a purpose, though it's used in two different ways. What would be considered abusing it, and why is that a problem?

Screen Shot 2022-08-29 at 16 20 52
Screen Shot 2022-08-29 at 16 21 10

app/assets/stylesheets/config/_colors.css
vasconsaurus: I think these classes help to be more flexible when applying our texts colors. `color-white` for me reads a bit confusing, as it could be the background-color, instead of just text? Though I know color refers to the text. Maybe `txt-color-white`?

app/assets/stylesheets/config/_spaces.css
vasconsaurus: Using the number 15/30/60 helped me understand a bit faster which one I was using, BUT I totally agree with you. I think on the long run this makes more sense. Thinking in terms of proportions will make it easier to change things if needed. 

I took a look at your file, and decided to go with s, m, and l for our sizes.

app/assets/stylesheets/components/_navbar.css
vasconsaurus: I had read about using rem with the base unit defined at the body before, I thought it was super interesting, also read the article you linked, so I went with that. (the old design habits kind of kicked in with the pt as units 😬 perdão)

app/assets/stylesheets/application.css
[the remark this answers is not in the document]
tcione: > I have not. Would I check this by checking the server? Could you maybe explain to me what I'm looking for?

The network tab. When using import in pure CSS, it makes an individual request for each `import`.

> Also, not sure if it's related, but what do you think about working with sass, adding the 'sassc-rails' gem?

Not related, but I'd prefer it, since it adds some convenience without having to overblow the project with stuff like webpacker.

app/assets/stylesheets/application.css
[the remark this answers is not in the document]
tcione: You don't have to comment and uncomment a line, then refresh the full page. You can simply add a class using the web inspector :D

app/assets/stylesheets/application.css
```diff
@@ -13,3 +13,32 @@
...
+}
+
+html, body { height: 100% }
+
+body {
+  background-color: var(--yellow);
+  display: flex;
+  flex-direction: column;
+}
```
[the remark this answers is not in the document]
tcione: By as is, you mean as it is handled by the browser? Then yes.  That said, it's a matter of preference, I do not like to attach rules that are not general to the body, so I restrict myself to font-size, color, etc.

app/assets/stylesheets/components/_navbar.css
tcione: All good, almost every one used to start using points and shifted later :p

app/assets/stylesheets/components/_navbar.css
tcione: Thinking back it's the old man in me speaking. I tend to prefer not adding styles to things when possible, so I usually think in this sequence:

1. Can I rely on the normalized behavior from `_base.css`?
2. Can I rely on the elements inheritance?
3. What's the minimum amount of rules to achieve what I'm trying to do here?

In this specific case you could:
- Mobile: Make children of the element `display: block`
- Desktop: Make children of the element `display: inline-block`

And nothing else to guarantee the positioning you want.

All that said, there is performance penalties, but they should be minimal, so I'm fine if you prefer things the way the are, this is mostly a matter of taste.

app/assets/stylesheets/config/_colors.css
tcione: Hm, maybe I'm just biased since I've been using this for a while. I'd be fine with just inverting the class names `txt-white` instead of `txt-yellow`.

In a sense, I like these utility classes reflecting their implementation meaning `padding-s` means `small padding property` and `color-white` means `white` `color` property.

.gitignore
```diff
@@ -31,3 +31,6 @@
 /config/master.key
 coverage/
 .env
+
+# Ignore DS Store
+**/.DS_Store
```
tcione, line 36: **suggestion:**: It's fine to have this here, but you could also ad `.DS_Store` to your global `.gitignore` (in your home folder).

That said, I think `.DS_Store` instead of `**/.DS_Store` would be enough.

app/assets/stylesheets/components/_button-holder.css
```diff
@@ -0,0 +1,8 @@
+.button-holder {
```
tcione, line 1: **question:** is this a button variation? If so, should we name it `button--holder`?

app/assets/stylesheets/components/_navbar.css
```diff
@@ -0,0 +1,52 @@
...
+  font-family: eds-market-bold-slant, sans-serif;
+  font-size: var(--font-size-s);
+  line-height: 1.25;
+  color: var(--blue);
+  text-align: center;
+}
+
+.design-elements {
+  font-family: Eds Market Design Elements;
+}
```
tcione, line 25: **question:** Are we using this class somewhere?

app/assets/stylesheets/config/_base.css
```diff
@@ -0,0 +1,48 @@
...
+  --font-size-s: 2rem;
+  --font-size-m: 2.5rem;
+  --font-size-l: 4rem;
+  --font-size-xl: 6.5rem;
+}
+
+*, *::before, *::after {
+  box-sizing: border-box;
+  margin: 0px;
+  padding: 0px;
```
tcione, line 15: **sugestion:** It's a matter of personal preference, but I like to omit units from `0` values

```suggestion
  margin: 0;
  padding: 0;
```

app/assets/stylesheets/config/_colors.css
```diff
@@ -0,0 +1,23 @@
+html {
+  --light-beige: #E8DBB3;
+  --yellow: #FFD96A;
+  --red: #D92A45;
+  --light-blue: #7387EB;
+  --blue: #2347EB;
```
tcione, line 6: **suggestion:** we should move these to `:root` in `_base`. WDYT? 

I also like to prefix variables with their type, so I'd do something like `--color--light-beiger`. But I'm fine with not doing that.

app/assets/stylesheets/config/_base.css
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/* body *, body *::before, body *::after {
+  border: 0.5px solid blue;
+} */
```
tcione, line 20:
```suggestion
.debug *, .debug *::before, .debug *::after {
  border: 0.5px solid blue;
}
```

app/assets/stylesheets/config/_base.css
```diff
@@ -0,0 +1,48 @@
...
+    /*
+  This allows us to use readable rems.
+  - 24px becomes 2.4rem
+  - 10px becomes 1rem
+  - etc
+   */
+  font-size: 62.5%;
+}
+
+html, body { height: 100% }
```
tcione, line 32: **question:** Why setting height?

app/assets/stylesheets/config/_fonts.css
```diff
@@ -0,0 +1,59 @@
...
+  font-size: var(--font-size-m);
+}
+
+h4, h5 {
+  font-size: var(--font-size-s);
+}
+
+h6 {
+  font-size: var(--font-size-xs);
+}
```
tcione, line 41: **suggestion:** should we move these to `base`?

app/assets/stylesheets/config/_rotations.css
```diff
@@ -0,0 +1,61 @@
+.rotate-zero {
+  transition:  0.5s ease-in-out;
```
tcione, line 2: **nitpick:**

```suggestion
  transition: 0.5s ease-in-out;
```

app/assets/stylesheets/config/_spaces.css
```diff
@@ -0,0 +1,64 @@
+/* paddings */
+.p-top-s {
+  padding-top: 15px;
```
tcione, line 3: **suggestion**: Should we extract these pixel sizes to variables? Also should we consider transforming these in `rem`?

app/assets/stylesheets/pages/homepage.css
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.banner-about {
+  background-color: var(--blue);
+  display: flex;
+  flex-flow: row nowrap;
+  align-items: center;
+  justify-content: space-between;
+  text-align: center;
+  gap: 30px;
```
tcione, line 35: **suggestion:** this looks consistent with paddings. We should consider extracting spacing values to variables.

app/assets/stylesheets/application.css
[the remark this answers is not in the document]
vasconsaurus: I looked at the network tab, but it looks like it's not requesting each one individually:

Screen Shot 2022-09-02 at 13 51 24

app/assets/stylesheets/application.css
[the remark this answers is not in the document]
vasconsaurus: uuuuuuuuuu! that's clever, would never think of it 😅

app/assets/stylesheets/application.css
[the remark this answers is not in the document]
vasconsaurus: Sorry, I think I didn't phrase it properly. But I get what you meant now. 

I think, considering our designs, the pages will have the same basic flow. So we probably won't need the extra div. Can we leave as it is for now, and add an extra div if needed?

app/assets/stylesheets/components/_navbar.css
vasconsaurus: I see, since I tend to over rely on flexbox, I changed it to use `block` and `inline-block`.
I did have, however to add some extra changes to get it to work.

desktop:
I had to add a `display: inline` to the p, and a paddings to the first link.

mobile:
I had to add `display: block` to everyone, `text-align: right` and `padding: 0` to the links.

app/assets/stylesheets/config/_colors.css
vasconsaurus: I changed to `color-white` and so on, it still feels weird to me, but I think it's a good way for me to get used to the fact that `color` refers to font color 🥲

app/assets/stylesheets/components/_button-holder.css
vasconsaurus: no, it really is a button holder, a div that holds the buttons 😬

app/assets/stylesheets/components/_navbar.css
vasconsaurus: no, I was going to, but didn't. I'm going to delete it.

app/assets/stylesheets/config/_colors.css
vasconsaurus: I moved before I saw this, so, yes, completely agree 😂

app/assets/stylesheets/config/_fonts.css
vasconsaurus: I thought of that, but maybe the base would get too bloated? I'm not sure, WDYT?

app/assets/stylesheets/config/_base.css
vasconsaurus: It was for something I tried before and forgot to delete 😬

app/assets/stylesheets/application.css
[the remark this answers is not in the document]
tcione: Then there's nothing to worry about :D

app/assets/stylesheets/config/_fonts.css
tcione: I think it depends on what makes more sense for your future self. Bloat or not is rarely the right question for me. In these cases, I tend to think in terms of what is easier for me to navigate. If you feel like fonts is better, then let's go for fonts. For me personally, if there is not a tag, I search in base haha

app/assets/stylesheets/components/_buttons.css
```diff
@@ -0,0 +1,13 @@
...
+  height: 120px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.button--yellow {
+  background-color: var(--color--yellow);
+  outline: 6px solid var(--color--red);
+  outline-offset: -10px;
```
tcione, line 12:
```suggestion
  outline: 0.6rem solid var(--color--red);
  outline-offset: -1rem;
```

app/assets/stylesheets/components/_navbar.css
```diff
@@ -0,0 +1,52 @@
...
+  justify-content: space-between;
+  align-items: center;
+  background-color: var(--color--yellow);
+  position: fixed;
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 1;
+  width: 100vw;
+  height: 100px;
```
tcione, line 12:
```suggestion
  width: 100vw;
  height: 10rem;
```

.gitignore
tcione, line 36:
```suggestion
.DS_Store
```

app/assets/stylesheets/config/_fonts.css
vasconsaurus: I think for me keeping it in a fonts file might be easier. I'll leave it like this for now and see how it works 🙂

vasconsaurus:
@tcione , I think now everything is done?

Alt Text

tcione:
@vasconsaurus yeap, pretty much :)

PR #48, opened 2022-09-12
Feature/45/item seasons/index/frontend

### LINKED ISSUE #45: feature - item seasons #index: frontend
- [ ] fix the footer position

vasconsaurus:
closes #45 

First version of our item seasons index frontend.

**On desktop**
Screen Shot 2022-09-12 at 13 26 47
Screen Shot 2022-09-12 at 13 26 51

**On mobile**
Screen Shot 2022-09-12 at 13 37 48
Screen Shot 2022-09-12 at 13 37 55

**On surface pro**
Screen Shot 2022-09-12 at 13 37 29

**Short video**
https://user-images.githubusercontent.com/87862340/189708242-ad2d5ea3-a68f-469e-858b-3b6dcd451672.mov

**Of note**
Added two new helpers. One in item seasons for the rotation of the buttons, and one in application for the abbreviated months. Also added tests for both.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -14,16 +14,15 @@
  *= require_self
  */
 
+@import "components/banner-small";
 @import "components/button-holder";
 @import "components/buttons";
+@import "components/content";
 @import "components/footer";
 @import "components/navbar";
 @import "config/base";
+@import "config/colors";
 @import "config/fonts";
 @import "config/rotations";
 @import "config/spaces";
 @import "pages/homepage.css";
-
-.content {
-  flex: 1 0 auto;
-}
--- a/app/assets/stylesheets/config/_base.css
+++ b/app/assets/stylesheets/config/_base.css
@@ -29,6 +29,10 @@
   border: 0.5px solid blue;
 }
 
+html, body {
+  height: 100%;
+}
+
 html {
     /*
   This allows us to use readable rems.
--- a/app/assets/stylesheets/config/_base.css
+++ b/app/assets/stylesheets/config/_base.css
@@ -15,6 +15,7 @@
   --spaces-size-s:1.5rem;
   --spaces-size-m:3rem;
   --spaces-size-l:6rem;
+  --size-nav:10rem;
 }
 
 *, *::before, *::after {
--- a/app/helpers/item_seasons_helper.rb
+++ b/app/helpers/item_seasons_helper.rb
@@ -2,7 +2,7 @@
 
 module ItemSeasonsHelper
   def sampled_rotation
-    ["rotate-zero", "rotate-plus-5", "rotate-plus-10", "rotate-plus-15",
-      "rotate-minus-5", "rotate-minus-10", "rotate-minus-15"].sample
+    %w[rotate-zero rotate-plus-5 rotate-plus-10 rotate-plus-15
+       rotate-minus-5 rotate-minus-10 rotate-minus-15].sample
   end
 end
--- a/app/helpers/item_seasons_helper.rb
+++ b/app/helpers/item_seasons_helper.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module ItemSeasonsHelper
-  def sampled_rotation
+  def randomized_rotation
     %w[rotate-zero rotate-plus-5 rotate-plus-10 rotate-plus-15
        rotate-minus-5 rotate-minus-10 rotate-minus-15].sample
   end
--- a/app/helpers/item_seasons_helper.rb
+++ b/app/helpers/item_seasons_helper.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module ItemSeasonsHelper
+  def sampled_rotation
+    ["rotate-zero", "rotate-plus-5", "rotate-plus-10", "rotate-plus-15",
+      "rotate-minus-5", "rotate-minus-10", "rotate-minus-15"].sample
+  end
+end
--- a/app/views/home_page/index.html.erb
+++ b/app/views/home_page/index.html.erb
@@ -1,27 +1,29 @@
-<section class="banner-hero p-top-l p-bottom-l p-sides-m">
-  <div class="script-md banner-hero__txt"><%= I18n.t('root.hero') %></div>
-</section>
+<div class="content">
+  <section class="banner-hero p-top-l p-bottom-l p-sides-m">
+    <div class="script-md banner-hero__txt"><%= I18n.t('root.hero') %></div>
+  </section>
 
-<section class="banner-buttons p-top-m p-bottom-l p-sides-m">
-  <div class="p-bottom-m">
-    <h3 class="color-yellow"><%= I18n.t('action.search_by') %></h3>
-  </div>
-  <div class="button-holder">
-      <%= link_to I18n.t('action.month').capitalize, item_seasons_path, class: 'link_action button button--yellow script-lg rotate-minus-5' %>
-      <%= link_to I18n.t('action.produce').capitalize, produce_items_path, class: 'link_action button button--yellow script-lg rotate-zero' %>
-  </div>
-</section>
-
-<section class="banner-about p-top-m p-bottom-l">
-  <%= image_tag "https://dm2305files.storage.live.com/y4mYqMHpvb-2aKz6Q7P3jqr0EugNasCWn476P-zPwozPAPAxnKwadlz3S7DUa3FvDB2XLQdEI8rOvSkZTcITqKiIurRon4Fywldfm-p6DTsaNLzfXqXZUkFrkUVwxOswKTPKuMkElCHTjJZdNxSLwkr7bXSW3HyRvrI6J2MOl813zCFrm9YT9yP0pPecIaR9LZY?width=285&height=1217&cropmode=none", class: 'banner-about-img img-left' %>
-  <div class="banner-about__txt">
+  <section class="banner-buttons p-top-m p-bottom-l p-sides-m">
     <div class="p-bottom-m">
-        <h3 class="color-white"><%= I18n.t('root.about_title') %></h3>
+      <h3 class="color-yellow"><%= I18n.t('action.search_by') %></h3>
     </div>
-    <div>
-      <p class="color-white p-bottom-m"><%= I18n.t('root.about_text_p1') %></p>
-      <p class="color-white"><%= I18n.t('root.about_text_p2') %></p>
+    <div class="button-holder">
+        <%= link_to I18n.t('action.month').capitalize, item_seasons_path, class: 'link_action button button--yellow script-lg rotate-minus-5' %>
+        <%= link_to I18n.t('action.produce').capitalize, produce_items_path, class: 'link_action button button--yellow script-lg rotate-zero' %>
     </div>
-  </div>
-  <%= image_tag "https://dm2305files.storage.live.com/y4m9pwZq1iIqcdRBSt8VZEwJN6qA1UvxiiNXGFPETpPL4z08lKRDfpGUFuQ7CsrpYE0Vpou1Xgz-Gz4LcEd2mTKZyWSlfjXPzvinyXGo9SgxZwpiKxf6HWX6ui_Bn0s0uFQAd9Qxvccnz0ulEzPGQNW9CwIdfJtYP_89wi8HPn6ZdpqE9zWtZWO3GOMoJAwJgVF?width=269&height=1217&cropmode=none", class: 'banner-about-img img-right' %>
-</section>
+  </section>
+
+  <section class="banner-about p-top-m p-bottom-l">
+    <%= image_tag "https://dm2305files.storage.live.com/y4mYqMHpvb-2aKz6Q7P3jqr0EugNasCWn476P-zPwozPAPAxnKwadlz3S7DUa3FvDB2XLQdEI8rOvSkZTcITqKiIurRon4Fywldfm-p6DTsaNLzfXqXZUkFrkUVwxOswKTPKuMkElCHTjJZdNxSLwkr7bXSW3HyRvrI6J2MOl813zCFrm9YT9yP0pPecIaR9LZY?width=285&height=1217&cropmode=none", class: 'banner-about-img img-left' %>
+    <div class="banner-about__txt">
+      <div class="p-bottom-m">
+          <h3 class="color-white"><%= I18n.t('root.about_title') %></h3>
+      </div>
+      <div>
+        <p class="color-white p-bottom-m"><%= I18n.t('root.about_text_p1') %></p>
+        <p class="color-white"><%= I18n.t('root.about_text_p2') %></p>
+      </div>
+    </div>
+    <%= image_tag "https://dm2305files.storage.live.com/y4m9pwZq1iIqcdRBSt8VZEwJN6qA1UvxiiNXGFPETpPL4z08lKRDfpGUFuQ7CsrpYE0Vpou1Xgz-Gz4LcEd2mTKZyWSlfjXPzvinyXGo9SgxZwpiKxf6HWX6ui_Bn0s0uFQAd9Qxvccnz0ulEzPGQNW9CwIdfJtYP_89wi8HPn6ZdpqE9zWtZWO3GOMoJAwJgVF?width=269&height=1217&cropmode=none", class: 'banner-about-img img-right' %>
+  </section>
+</div>
--- a/app/views/item_seasons/index.html.erb
+++ b/app/views/item_seasons/index.html.erb
@@ -1,5 +1,5 @@
 <div class="content background-color-blue">
-  <h1 class="page__title banner-small"> <%= I18n.t("item_season.index.title") %> </h1>
+  <h1 class="page__title banner-small p-top-m p-bottom-m"> <%= I18n.t("item_season.index.title") %> </h1>
 
   <div class="button-holder p-top-l p-bottom-l">
     <% ItemSeason::MONTHS.each do |month_index| %>
--- a/app/views/item_seasons/index.html.erb
+++ b/app/views/item_seasons/index.html.erb
@@ -1,9 +1,9 @@
-<h1 class="page__title"> <%= I18n.t("item_season.index.title") %> </h1>
+<div class="content background-color-blue">
+  <h1 class="page__title banner-small"> <%= I18n.t("item_season.index.title") %> </h1>
 
-<table class="produce-table">
-  <% ItemSeason::MONTHS.each do |month_index| %>
-    <tr>
-      <td class="produce-table__month"><%= link_to (formatted_month_name(month_index)).capitalize, item_season_path(month_index), class: "link_action month_link" %></td>
-    </tr>
-  <%  end %>
-</table>
+  <div class="button-holder p-top-l p-bottom-l">
+    <% ItemSeason::MONTHS.each do |month_index| %>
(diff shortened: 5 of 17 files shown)


app/assets/stylesheets/application.css
```diff
@@ -14,16 +14,15 @@
...
 @import "components/buttons";
+@import "components/content";
 @import "components/footer";
 @import "components/navbar";
 @import "config/base";
+@import "config/colors";
 @import "config/fonts";
 @import "config/rotations";
 @import "config/spaces";
 @import "pages/homepage.css";
```
tcione, line 28: **nitpick**: I usually prefer to organize imports from less specific to more specific in order to enact proper cascading. In this case we'd start from config, to components and then to pages.

app/assets/stylesheets/config/_base.css
```diff
@@ -28,6 +29,10 @@
   border: 0.5px solid blue;
 }
 
+html, body {
+  height: 100%;
+}
```
tcione, line 34: **question:** why is this needed?

app/views/home_page/index.html.erb
```diff
@@ -1,27 +1,29 @@
...
-  <%= image_tag "https://dm2305files.storage.live.com/y4mYqMHpvb-2aKz6Q7P3jqr0EugNasCWn476P-zPwozPAPAxnKwadlz3S7DUa3FvDB2XLQdEI8rOvSkZTcITqKiIurRon4Fywldfm-p6DTsaNLzfXqXZUkFrkUVwxOswKTPKuMkElCHTjJZdNxSLwkr7bXSW3HyRvrI6J2MOl813zCFrm9YT9yP0pPecIaR9LZY?width=285&height=1217&cropmode=none", class: 'banner-about-img img-left' %>
-  <div class="banner-about__txt">
+  <section class="banner-buttons p-top-m p-bottom-l p-sides-m">
     <div class="p-bottom-m">
-        <h3 class="color-white"><%= I18n.t('root.about_title') %></h3>
+      <h3 class="color-yellow"><%= I18n.t('action.search_by') %></h3>
+    </div>
+    <div class="button-holder">
+        <%= link_to I18n.t('action.month').capitalize, item_seasons_path, class: 'link_action button button--yellow script-lg rotate-minus-5' %>
+        <%= link_to I18n.t('action.produce').capitalize, produce_items_path, class: 'link_action button button--yellow script-lg rotate-zero' %>
```
tcione, line 12: **question:** why `link_to` over `normal `<a />`?

spec/helpers/application_helper_spec.rb
```diff
@@ -18,4 +18,10 @@
       expect(helper.formatted_month_name(1)).to eq('janeiro')
     end
   end
+
+  describe 'short month name' do
```
tcione, line 22: **suggestion:** I normally prefer to describe method tests with the method name. In this case:

```suggestion
  describe '#formatted_month_short' do
```

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ItemSeasonsHelper, type: :helper do
+  describe 'randomized rotation' do
+    it 'adds a random rotation to a button' do
+      expect(helper.randomized_rotation).to include('rotate')
+    end
+  end
```
tcione, line 10: **issue:** you are not testing much here. All items have rotate, thus you just know it return a member of the array. Testing randomness is tricky and probably a bit flaky. So I'd suggest to do something like this:

```rb
array1 = 10.times.map { helper.randomized_rotation }
array2 = 10.times.map { helper.randomized_rotation }
expect(array1).not_to match(array2)
```

app/views/item_seasons/index.html.erb
```diff
@@ -1,9 +1,10 @@
...
-  <% ItemSeason::MONTHS.each do |month_index| %>
-    <tr>
-      <td class="produce-table__month"><%= link_to (formatted_month_name(month_index)).capitalize, item_season_path(month_index), class: "link_action month_link" %></td>
-    </tr>
-  <%  end %>
-</table>
+  <div class="button-holder p-top-l p-bottom-l p-sides-s">
+    <% ItemSeason::MONTHS.each do |month_index| %>
+      <%= link_to (formatted_month_short(month_index)).capitalize, item_season_path(month_index),
+          class: "link_action month_link button button--yellow script-lg m-top-s m-bottom-s #{randomized_rotation}" %>
```
tcione, line 7: **nitpick:** this is not relevant at this scale, so feel free to ignore this comment. I'm leaving it here just so you are generally aware of this type of issue.

Here you are calculating something random 12 times. Not sure how Ruby handles this and I'm sure the performance hit is negligible even in cases that you calculate this 100mi times. That said, it's always worthwhile to consider these questions (like, what's the impact of calculating a value on every loop iteration).

A more performance approach would be doing some deterministic math using the month index. But, then I don't think it's worth your time to worry about an alternative here.

tcione (commented):
Great work here. Also like the layout <3. Left a few minor comments for your consideration.

app/assets/stylesheets/config/_base.css
vasconsaurus: I used this logic for the footer to stay in the bottom:
```
<body>
  <div class="content">
    content
  </div>
  <footer class="footer"></footer>
</body>

html, body {
  height: 100%;
}
body {
  display: flex;
  flex-direction: column;
}
.content {
  flex: 1 0 auto;
}
.footer {
  flex-shrink: 0;
}
```

Without the set height it doesn't work. I admit though that I don't quite understand why it doesn't work without it. I need to read a bit more on it.

app/views/home_page/index.html.erb
vasconsaurus: what do you mean by `normal`?

app/views/home_page/index.html.erb
```diff
@@ -1,27 +1,29 @@
...
     </div>
-    <div>
-      <p class="color-white p-bottom-m"><%= I18n.t('root.about_text_p1') %></p>
-      <p class="color-white"><%= I18n.t('root.about_text_p2') %></p>
+  </section>
+
+  <section class="banner-about p-top-m p-bottom-l">
+    <%= image_tag "https://dm2305files.storage.live.com/y4mYqMHpvb-2aKz6Q7P3jqr0EugNasCWn476P-zPwozPAPAxnKwadlz3S7DUa3FvDB2XLQdEI8rOvSkZTcITqKiIurRon4Fywldfm-p6DTsaNLzfXqXZUkFrkUVwxOswKTPKuMkElCHTjJZdNxSLwkr7bXSW3HyRvrI6J2MOl813zCFrm9YT9yP0pPecIaR9LZY?width=285&height=1217&cropmode=none", class: 'banner-about-img img-left' %>
```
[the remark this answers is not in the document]
vasconsaurus: I wanted to link the image from assets, but couldn't get it to work. I'm linking from an external URL while that doesn't work.

spec/helpers/item_seasons_helper_spec.rb
vasconsaurus: Yeah, I had a hard time trying to figure out how I could test this one. I thought I could at least test if the rotation was being added.

app/views/home_page/index.html.erb
tcione: oh, sorry, I messed up the description. I meant a normal ` <a />`

app/views/item_seasons/index.html.erb
vasconsaurus: At first, when designing the layout, I thought I would make specific rotations for specific months. But while working I thought the random rotations might be interesting. I think always changing is interesting, but it can look uglier, and there is the performance side of things.

So i made a version with the 'fixed' rotations. WDYT?

app/views/home_page/index.html.erb
vasconsaurus: I was taught to use them as they were easier to maintain. Though that can be just a teachers opinion.

However, if I' m going to end using them in some more complicated instances, I find it easier to always use them. Instead of going between different ways to write the links.

tcione: Sure, that's purely a matter of taste, I avoid any html helpers (and type the tags manually with whatever attributes would be rendered). Anyhow, please disregard my comment.

app/views/item_seasons/index.html.erb
tcione: Sure, makes sense. The original way was fine as well, I just took the opportunity to expose aspects you should consider in more complex cases (as of indicating a certain mindset to have).

spec/helpers/application_helper_spec.rb
```diff
@@ -2,20 +2,16 @@
...
-#   end
-# end
 RSpec.describe ApplicationHelper, type: :helper do
   describe 'formatted month name' do
     it 'changes the month_index to a month_name' do
       expect(helper.formatted_month_name(1)).to eq('janeiro')
     end
   end
+
+  describe 'formatted month short' do
```
tcione, line 12: **suggestion:** I meant literally `#formatted_month_short`, since you are describing the method. In Ruby notation, `#` indicates instance methods and `.` or `::` indicates class methods. Using the following as example:

```rb
class MyExampleClass
  def self.a_class_method; end
  
  def an_instance_method; end
end
```

We would use the following to refer to them:
`MyExampleClass.a_class_method` or `MyExampleClass::a_class_method`
`MyExampleClass#an_instance_method`

You can see a clear example in the [Ruby docs](https://ruby-doc.org/core-3.1.0/Array.html). For that reason, I prefer to follow that convention in tests, after all you are describing a method.

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ItemSeasonsHelper, type: :helper do
+  describe 'rotation' do
+    it 'adds rotation to month with index 12' do
+      expect(helper.rotation(12)).to eq('rotate-plus-10')
+    end
```
tcione, line 9: **issue**: We should cover all cases, here you are just verifying that `12` yields the expected output. If we refactored the method to
```rb
def rotation(index)
  return 'rotate-plus-10' if index == 12
  
  'rotate-zero'
end
```

Tests would still pass.

tcione (changes requested):

app/helpers/item_seasons_helper.rb
```diff
@@ -0,0 +1,18 @@
...
+    when 1, 6
+      'rotate-plus-5'
+    when 5
+      'rotate-minus-5'
+    when 12
+      'rotate-plus-10'
+    when 4, 9
+      'rotate-minus-10'
+    else
+      'rotate-zero'
```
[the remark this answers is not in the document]
vasconsaurus: because I like writing `zero` 😅. I'll change it to `0` for consistency.

spec/helpers/application_helper_spec.rb
vasconsaurus: Ok! Then I'll also change `formatted month name` to `#formatted_month_name`. (I'll also checkout the docs)

app/views/item_seasons/index.html.erb
vasconsaurus: I think it is a great exercise in weighting pros and cons of different solutions, and deciding what makes more sense. To think beyond our first solution. I think I'm going with our second solution, I think the greater control of what gets rendered is nice.

spec/helpers/item_seasons_helper_spec.rb
vasconsaurus: I added tests for all the rotations

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+require 'rails_helper'
+
+RSpec.describe ItemSeasonsHelper, type: :helper do
+  describe '#rotation' do
+    it 'adds rotation +10 to month with index 12' do
+      expect(helper.rotation(12)).to eq('rotate-plus-10')
+    end
+
+    it 'adds rotation +5 to month with index 1 and 6' do
+      expect(helper.rotation(1 || 6)).to eq('rotate-plus-5')
```
tcione, line 12: **issue:** `1 || 6` will always return `1`. We need one case per integer

tcione (changes requested):

spec/helpers/item_seasons_helper_spec.rb
tcione: You could use test tables:
```rb
[1, 6].each do |number|
  it "adds rotation +5 to month with index #{number}" do
    expect(helper.rotation(number)).to eq('rotate-plus-5')
  end
end
```

This is not a very common pattern in ruby (comes from `go`), but it's quite useful. We could even go further and:

```rb
[
  { number: 1, rotation: 'plus-5' },
  { number: 6, rotation: 'plus-5' },
  # all other possibilities
].each do |example|
  it "adds rotation #{example[:rotation]} to month with index #{example[:number]}" do
    expect(helper.rotation(number)).to eq("rotate-#{example[:rotation]}")
  end
end
```

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ItemSeasonsHelper, type: :helper do
+  describe '#rotation' do
```
tcione, line 6: **question:** What happens when we give an unexpected number (let's say 52). What would happen? Should we add a test case for this?

spec/helpers/item_seasons_helper_spec.rb
vasconsaurus: > **issue:** `1 || 6` will always return `1`. We need one case per integer

derp! true! Sometimes the simple things still get me 😬

spec/helpers/item_seasons_helper_spec.rb
vasconsaurus: Hum, good question!

```
<div class="button-holder p-top-l p-bottom-l p-sides-s">
    <% ItemSeason::MONTHS.each do |month_index| %>
      <%= link_to (formatted_month_short(month_index)).capitalize, item_season_path(month_index),
          class: "link_action month_link button button--yellow script-lg m-top-s m-bottom-s #{rotation(month_index)}" %>
    <%  end %>
  </div>
```

This is what we have right now. If we got a 52-month-season in the rotation, it would rotate to 0. However, we don't want an extra month to show at all. So I think there is a bigger problem here.

I made some tests and I think a way of dealing with it might be only running if `month_index <= 12`. This way we make sure we are only ever showing 12 months.

```
  <div class="button-holder p-top-l p-bottom-l p-sides-s">
    <% ItemSeason::MONTHS.each do |month_index| %>
      <% if month_index <= 12 %>
        <%= link_to (formatted_month_short(month_index)), item_season_path(month_index),
              class: "link_action month_link button button--yellow script-lg m-top-s m-bottom-s #{rotation(month_index)}" %>
      <%  end %>
    <% end %>
  </div>
```

Then we could maybe add a test case to views or features to check if we only have 12 months.

WDYT?

tcione: You went way deeper than what I expected :)

### About my question per se

I was talking  purely from a code perspective. The function knows nothing of your restrictions. So from a functional (and testing) perspective we should make sure that when fed with an unexpected number, it returns something you expect (in this case, `0` rotation). 

So my objective was just making you a test case that covered this.

### About where the question branched to

There's no need to do a check, you control the constant. If there are more numbers there, then that should be dealt directly in the constant. You just need to add guards and all when you do not control the data (i.e: it comes from user input).

vasconsaurus: I added the test and removed the check 🙂 

But I have a question: even if we don't want this to run at all, do we still need to do the test? Because, if it goes through we expect it to return zero, but in the end, it shouldn't get to that point. So I'm confused about why should we test for something that shouldn't happen.

tcione: Because even though it shouldn't, the code allows for it. Assuming you work in a bigger team, anyone could misuse the method. There's nothing there explicitly saying you expect numbers from 1 to 12.

That said, i think my point was misguided, there are some intervals between 1 and 12 that do not have a specific rotation, so this case is already covered :)

Sorry for that

tcione: But to answering the "if we don't want this to run at all" comment: if we don't want it to run outside of expected parameters, the best thing to do would be to raise an error (i.e: `raise 'index must be between 1 and 12`]

vasconsaurus: Added the error to the helper and to the test. Though I'm not sure if the test I wrote is quite right. I wanted to test for any number bigger than 12, so I used rand, but I had to use a max value.

app/helpers/item_seasons_helper.rb
```diff
@@ -0,0 +1,20 @@
...
+    when 12
+      'rotate-plus-10'
+    when 4, 9
+      'rotate-minus-10'
+    when 2, 3, 7, 8, 10, 11
+      'rotate-0'
+    else
+      raise 'month-index must be between 1 and 12'
+    end
+  end
```
tcione, line 19: **suggestion:** I think using guard clauses would be easier to read:

```suggestion
  def rotation(month_index)
    raise 'month-index must be between 1 and 12' unless 1..12.include?(month_index)
    
    case month_index
    when 1, 6
      'rotate-plus-5'
    when 5
      'rotate-minus-5'
    when 12
      'rotate-plus-10'
    when 4, 9
      'rotate-minus-10'
    else
      'rotate-0'
    end
  end
```

spec/helpers/item_seasons_helper_spec.rb
tcione: You should test for numbers smaller than 1 and bigger than 12. That said, you do not need to get a random number, just the first one.

app/helpers/item_seasons_helper.rb
vasconsaurus: I accepted this change, but it wasn't passing the tests, and rubocop didn't like it. So I changed it.
Question: what is the benefit of using a guard clause here instead of how it was written before? Is the fact that it runs before anything the advantage?

tcione: On benefit: it's easier to read IMO. Putting what you expect as arguments as the very first thing sets clear rules for arguments. Performance-wise there is a bit of difference but it's negligible.

Now on Rubocop and test: what breaks and how does it break?

vasconsaurus: As it was written before: `raise 'month-index must be between 1 and 12' unless 1..12.include?(month_index)`

All the helper tests didn't pass, except for the last ones:

Screen Shot 2022-09-19 at 15 23 42

As for rubocop, we got: `Avoid the use of flip-flop operators`

But changing it to `raise 'month-index must be between 1 and 12' if month_index > 12 || month_index < 1` seems to have fixed our problems. Is this ok?

vasconsaurus: aaaaaaaah! Now I get it 😄 . I was wondering why it wasn't working, I didn't know what a flip flop operator was, so I rewrote it 😬 Since it's only a matter of taste, I'm leaving `raise 'month-index must be between 1 and 12' if month_index > 12 || month_index < 1`

vasconsaurus:
@tcione, question: it says that there is one pending change request. But I think we are all done? Partiu?

tcione (approved):
LGTM

PR #50, opened 2022-09-20
Add layout to item-seasons show

### LINKED ISSUE #46: feature - item seasons #show: frontend

vasconsaurus:
closes #46 

The front end for our item seasons show. It doesn't have the sort feature yet, but I thought maybe we should approve the basic front end first, and then implement the sort. WDYT?

**On desktop**
Screen Shot 2022-09-20 at 10 26 50

**On mobile**
Screen Shot 2022-09-20 at 10 30 02

**Short video**
https://user-images.githubusercontent.com/87862340/191271496-61328343-3d72-4b63-ad3f-9e7202ca4688.mov


app/assets/stylesheets/application.css
```diff
@@ -14,6 +14,15 @@
  *= require_self
  */
 
+<<<<<<< HEAD
```
tcione, line 17: **issue:** we must remove the merge conflict marks
```
<<<<<<< HEAD
=======
>>>>>>> main
```

app/views/item_seasons/show.html.erb
```diff
@@ -1,13 +1,28 @@
...
-        <td class="produce-table__name"><%= I18n.t("produce_items.names.#{produce.name}").capitalize %></td>
-        <td class="produce-table__category"><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
-        <td class="produce-table__link_action"><%= link_to I18n.t('action.show'), produce, class: "link_action" %></td>
-    </tr>
-  <% end %>
-</table>
+  <div class="nav-actions-holder p-sides-m">
+    <h4><%= link_to I18n.t("action.back").upcase, item_seasons_path, class: "link_action back" %></h4>
+    <div class="nav-actions-months">
+      <%= link_to item_season_path(@item_season.month_index == 1 ? 12 : @item_season.month_index - 1) do %>
```
tcione, line 7: **suggestion:** extract `@item_season.month_index == 1 ? 12 : @item_season.month_index - 1` to a helper with a meaningful name. I.e: `previous_month_index`

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,7 @@
     <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
     <%= javascript_importmap_tags %>
     <link rel="stylesheet" href="https://use.typekit.net/mxd1nlu.css">
+    <script src="https://kit.fontawesome.com/e85e0c4e55.js" crossorigin="anonymous"></script>
```
tcione, line 12: **issue:** adding scripts to `<head />` block the page rendering. We should either move it to right before `</body>` or use the [`async` attribute](https://javascript.info/script-async-defer)

app/views/produce_items/index.html.erb
```diff
@@ -1,7 +1,7 @@
 <h1 class="page__title"> <%= I18n.t('produce_item.index.title') %> </h1>
 
 <table class="produce-table">
-    <% @produce_items.find_each do |produce|  %>
+    <% @produce_items.each do |produce|  %>
```
tcione, line 4: **question:** why change from `find_each` to `each`?

spec/views/item_seasons/show.html.erb_spec.rb
```diff
@@ -21,6 +21,6 @@
 
   it 'displays the produces' do
     render
-    expect(rendered).to match(/Carambola/)
+    expect(rendered).to match(/carambola/)
```
tcione, line 24: **suggestion:** If we don't care about casing, we could use the `i` modifier:

https://rubular.com/r/OJoM3T6RfTN108

So it'd become `/carambola/i`

tcione (changes requested):
Nicely done :D. Things I like in the PR:
✨ The description is quite good (I love the screenshots and videos)
✨ Layout looks good and general quality of code is good as well
✨ I like the approach of leaving sort to a later PR 

That said I have a few suggestions and issues left as comments and I'd like to propose changing the title from the branch name to something more meaningful, like "Add layout for xyz"

app/assets/stylesheets/application.css
vasconsaurus: I missed this one 😬

app/views/produce_items/index.html.erb
vasconsaurus: Sorry, I was experimenting with the sort and left this behind. If I'm not mistaken, I couldn't sort alphabetically with the find_each.

app/views/item_seasons/show.html.erb
```diff
@@ -1,13 +1,28 @@
...
-        <td class="produce-table__name"><%= I18n.t("produce_items.names.#{produce.name}").capitalize %></td>
-        <td class="produce-table__category"><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
-        <td class="produce-table__link_action"><%= link_to I18n.t('action.show'), produce, class: "link_action" %></td>
-    </tr>
-  <% end %>
-</table>
+  <div class="nav-actions-holder p-sides-m">
+    <h4><%= link_to I18n.t("action.back").upcase, item_seasons_path, class: "link_action back" %></h4>
+    <div class="nav-actions-months">
+      <%= link_to item_season_path(month_navigation_previous(@item_season)) do %>
```
tcione, line 7: **suggestion:** up to taste, but the whole thing could be moved to the helper. Ie.:

```rb
  def month_navigation_previous(item_season)
    index = item_season.month_index == 1 ? 12 : item_season.month_index - 1
    item_season_path(index)
  end
```

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -3,6 +3,10 @@
 require 'rails_helper'
 
 RSpec.describe ItemSeasonsHelper, type: :helper do
+  let(:item_season_jan) { ItemSeason.create(produce_item_id: 1, month_index: 1, country_code: 'br') }
+  let(:item_season_feb) { ItemSeason.create(produce_item_id: 1, month_index: 2, country_code: 'br') }
+  let(:item_season_dec) { ItemSeason.create(produce_item_id: 1, month_index: 12, country_code: 'br') }
```
tcione, line 8: **issue:** we should move this closer to where we need it. A suggestion of mine would be create a context:
```rb
context 'when navigating between seasons' do
  let(:item_season_jan) { ... }

  describe '#month_navigation_previous'

  describe '#month_navigation_next'

  # etc
end
```

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -3,6 +3,10 @@
 require 'rails_helper'
 
 RSpec.describe ItemSeasonsHelper, type: :helper do
+  let(:item_season_jan) { ItemSeason.create(produce_item_id: 1, month_index: 1, country_code: 'br') }
```
tcione, line 6: **issue (not for this PR):** Seeing `produce_item_id` makes me wonder if a foreign key was created. I would expect this to fail on creation, yet it doesn't.

I've created an issue to address it: https://github.com/vasconsaurus/produce_in_season/issues/51

**question:** Any reason for not using factories here?

**suggestion:** Building the item (using `new` instead of `create`) wouldn't persist the item in the DB. This would made tests faster and is a good practice in the long run (as the codebase grows, time it takes to run the test suite becomes a problem)

app/views/produce_items/index.html.erb
tcione: That is correct. This might become a problem as the database of items grows larger, but lets keep `each` for the time being, since it makes no different and alphabetic order makes more sense.

tcione: (disclosure why changing ordering breaks `find_each`: it is because find_each relies on IDs for batching records)

app/views/layouts/application.html.erb
vasconsaurus: makes sense 😅

spec/helpers/item_seasons_helper_spec.rb
vasconsaurus: > question: Any reason for not using factories here?

Not really, it was easier to write that way 😅  I changed it to use a factory. WDYT?

> issue (not for this PR): Seeing produce_item_id makes me wonder if a foreign key was created. I would expect this to fail on creation, yet it doesn't.

Can you maybe explain this issue? Or could we talk about it later, so I understand it better?

tcione:
> @tcione I tried to change the branch name, but I don't know how to associate it with this PR. Could you help me with that?

@vasconsaurus No need to change the branch name, just the PR title inside github :)

vasconsaurus:
> > @tcione I tried to change the branch name, but I don't know how to associate it with this PR. Could you help me with that?
> 
> @vasconsaurus No need to change the branch name, just the PR title inside github :)

gaaaah! I misunderstood 😆

spec/views/item_seasons/show.html.erb_spec.rb
tcione: Have you considered this one? :)

spec/helpers/item_seasons_helper_spec.rb
tcione: About foreign keys: in simplified terms they are links from one table to the other (in the database level). For instance, if you try to delete a produce item that has item seasons, PostgreSQL will complain saying that there are seasons attached to that item. As a alternative, you can also instruct the database software (RDMS – relational database management system) to do X or Y when they need to handle these cases.

I do recommend reading [this](https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-foreign-key/), which will provide you a better explanation on what foreign keys are and why use them :p

spec/factories/item_season.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :item_season do
+    produce_item_id { 1 }
```
tcione, line 5: **issue:** instead of a prefixed ID, we should add a produce item. WDYT?

spec/helpers/item_seasons_helper_spec.rb
```diff
@@ -30,4 +30,30 @@
...
+        expect(helper.month_navigation_previous(item_season_jan)).to eq(12)
+      end
+    end
+
+    describe '#month_navigation_next' do
+      it 'goes to the next month if item_season.month_index is smaller than 12' do
+        expect(helper.month_navigation_next(item_season_feb)).to eq(3)
+      end
+
+      it 'goes to the january if item_season.month_index is 12' do
```
tcione, line 54: **nitpick:** consistency in terminology. I think we should either say `1` and `12` or `jan` and `dec`. WDYT?

spec/helpers/item_seasons_helper_spec.rb
vasconsaurus: I know what they are, but I will read the article. I was confused because I thought we were using them. Clearly, I must have set up something wrong 😓

tcione: Aaah, sorry for that. I missed the foreign key reference [here](https://github.com/vasconsaurus/produce_in_season/blob/e93202d63b49813e24bc7d8d5971cb15c3126622/db/schema.rb#L34). My bad.

You shouldn't be able to create things with an arbitrary id, tho. I'll investigate this later.

spec/views/item_seasons/show.html.erb_spec.rb
vasconsaurus: whoops! forgot about this one 😅 added the`i` modifier

vasconsaurus:
@tcione now I think we should be good to go 🙂 WDYT?

spec/factories/item_season.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :item_season do
+    produce_item_id { build(:produce_item) }
```
tcione, line 5: **suggestion:** I'm kind of surprised FactoryBot is smart enough to fetch the id here, since you are building an entity, not accessing an id. A more explicit way to do that would be:
```suggestion
    produce_item { build(:produce_item) }
```

tcione (approved):
LGTM. Left a small suggestion, but up to you apply it or not :D

PR #52, opened 2022-10-17
Add layout to produce items/index

### LINKED ISSUE #47: feature - produce items #index: frontend

vasconsaurus:
Closes #47 

The first version of our produce items index front-end.

**On desktop**
Screen Shot 2022-10-17 at 14 19 05

**On mobile**
Screen Shot 2022-10-17 at 14 19 22

**Short video**
https://user-images.githubusercontent.com/87862340/196242632-0ebe5dde-6786-4470-ae9f-bb8246230481.mov

### DIFF
--- a/app/views/item_seasons/show.html.erb
+++ b/app/views/item_seasons/show.html.erb
@@ -18,7 +18,7 @@
     <% @item_season_produces.find_each do |produce| %>
       <div class="card">
         <div class="card-inner">
-          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action produce-name"  %></h4>
+          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action"  %></h4>
         </div>
           <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
       </div>
--- a/app/views/item_seasons/show.html.erb
+++ b/app/views/item_seasons/show.html.erb
@@ -18,7 +18,7 @@
     <% @item_season_produces.find_each do |produce| %>
       <div class="card">
         <div class="card-inner">
-          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action"  %></h4>
+          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action produce-name"  %></h4>
         </div>
           <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
       </div>
--- a/app/views/item_seasons/show.html.erb
+++ b/app/views/item_seasons/show.html.erb
@@ -1,13 +1,28 @@
-<h1 class="page__title"> <%= formatted_month_name(@item_season.month_index).capitalize %> </h1>
+<div class="content background-color-blue">
+  <h1 class="page__title banner-small m-bottom-m"> <%= formatted_month_name(@item_season.month_index).capitalize %> </h1>
 
-<table class="produce-table">
-  <% @item_season_produces.find_each do |produce| %>
-    <tr>
-        <td class="produce-table__name"><%= I18n.t("produce_items.names.#{produce.name}").capitalize %></td>
-        <td class="produce-table__category"><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
-        <td class="produce-table__link_action"><%= link_to I18n.t('action.show'), produce, class: "link_action" %></td>
-    </tr>
-  <% end %>
-</table>
+  <div class="nav-actions-holder p-sides-m">
+    <h4><%= link_to I18n.t("action.back").upcase, item_seasons_path, class: "link_action back" %></h4>
+    <div class="nav-actions-months">
+      <%= link_to item_season_path(month_navigation_previous(@item_season)) do %>
+        <i class="fa-solid fa-arrow-left-long" title="arrow-navigate-left" aria-hidden="true" aria-label="arrow-navigate-left"></i>
+      <% end %>
+      <h4 class="color-yellow p-sides-s"><%= I18n.t("action.month") %></h4>
+      <%= link_to item_season_path(month_navigation_next(@item_season)) do %>
+        <i class="fa-solid fa-arrow-right-long" title="arrow-navigate-right" aria-hidden="true" aria-label="arrow-navigate-right"></i>
+      <% end %>
+    </div>
+  </div>
 
-<%= link_to I18n.t("action.back"), :back, class: "link_action back" %>
+  <div class="card-holder p-top-l p-bottom-l p-sides-s">
+    <% @item_season_produces.find_each do |produce| %>
+      <div class="card">
+        <div class="card-inner">
+          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action"  %></h4>
+        </div>
+          <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
+      </div>
+    <% end %>
+  </div>
+
+</div>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -10,9 +10,9 @@
       <div class="card">
         <div class="card-inner">
           <h4 class="produce-name"><%= I18n.t("produce_items.names.#{produce.name}") %></h4>
-          <div>
+          <div class="card-inner__month-name">
             <%= produce.item_seasons.find_each do |item_season| %>
-              <%= formatted_month_name(item_season.month_index) %> /
+              <%= formatted_month_list(produce, item_season) %>
             <% end %>
           </div>
         </div>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -9,7 +9,12 @@
     <% @produce_items.find_each do |produce| %>
       <div class="card">
         <div class="card-inner">
-          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action produce-name"  %></h4>
+          <h4 class="produce-name"><%= I18n.t("produce_items.names.#{produce.name}") %></h4>
+          <div>
+            <%= produce.item_seasons.find_each do |item_season| %>
+              <%= formatted_month_name(item_season.month_index) %> /
+            <% end %>
+          </div>
         </div>
           <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
       </div>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -11,9 +11,7 @@
         <div class="card-inner">
           <h4 class="produce-name"><%= I18n.t("produce_items.names.#{produce.name}") %></h4>
           <div class="card-inner__month-name">
-            <%= produce.item_seasons.find_each do |item_season| %>
-              <%= formatted_month_list(produce, item_season) %>
-            <% end %>
+            <%= formatted_month_list(produce) %>
           </div>
         </div>
           <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -9,7 +9,7 @@
     <% @produce_items.find_each do |produce| %>
       <div class="card">
         <div class="card-inner">
-          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action"  %></h4>
+          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action produce-name"  %></h4>
         </div>
           <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
       </div>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -1,7 +1,7 @@
 <h1 class="page__title"> <%= I18n.t('produce_item.index.title') %> </h1>
 
 <table class="produce-table">
-    <% @produce_items.find_each do |produce|  %>
+    <% @produce_items.each do |produce|  %>
       <tr>
         <td class="produce-table__name"><%= I18n.t("produce_items.names.#{produce.name}").capitalize %></td>
         <td class="produce-table__category"><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -1,11 +1,18 @@
-<h1 class="page__title"> <%= I18n.t('produce_item.index.title') %> </h1>
+<div class="content background-color-blue">
+  <h1 class="page__title banner-small m-bottom-m"> <%= I18n.t('produce_item.index.title') %> </h1>
 
-<table class="produce-table">
-    <% @produce_items.each do |produce|  %>
-      <tr>
-        <td class="produce-table__name"><%= I18n.t("produce_items.names.#{produce.name}").capitalize %></td>
-        <td class="produce-table__category"><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
-        <td class="produce-table__link_action"><%= link_to I18n.t('action.show'), produce, class: "link_action produce_link" %></td>
-      </tr>
+  <div class="nav-actions-holder p-sides-m">
+    <h4><%= link_to I18n.t("action.back").upcase, root_path, class: "link_action back" %></h4>
+  </div>
+
+  <div class="card-holder p-top-l p-bottom-l p-sides-s">
+    <% @produce_items.find_each do |produce| %>
+      <div class="card">
+        <div class="card-inner">
+          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action"  %></h4>
+        </div>
+          <div class="produce-category"><%= I18n.t("produce_items.categories.#{produce.category}") %></div>
+      </div>
     <% end %>
-</table>
+  </div>
+</div>
--- a/app/views/produce_items/index.html.erb
+++ b/app/views/produce_items/index.html.erb
@@ -5,7 +5,7 @@
       <tr>
         <td class="produce-table__name"><%= I18n.t("produce_items.names.#{produce.name}").capitalize %></td>
         <td class="produce-table__category"><%= I18n.t("produce_items.categories.#{produce.category}") %></td>
-        <td class="produce-table__link_action"><%= link_to I18n.t('action.show'), produce, class: "link_action" %></td>
+        <td class="produce-table__link_action"><%= link_to I18n.t('action.show'), produce, class: "link_action produce_link" %></td>
       </tr>
     <% end %>
 </table>
--- a/libpq.5.dylib
+++ b/libpq.5.dylib
@@ -1,0 +1,1 @@
+/opt/homebrew/opt/postgresql@14/lib/postgresql@14/libpq.5.dylib/usr/local/lib/libpq.5.dylib
--- a/libpq.5.dylib
+++ b/libpq.5.dylib
@@ -1,1 +1,0 @@
-/opt/homebrew/opt/postgresql@14/lib/postgresql@14/libpq.5.dylib/usr/local/lib/libpq.5.dylib
--- a/spec/features/item_seasons_spec.rb
+++ b/spec/features/item_seasons_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe 'Item Seasons', type: :feature do
       :produce_item,
       name: 'morango',
       category: 'fruit',
-      item_seasons: [create(:item_season)]
+      item_seasons: [create(:item_season, produce_item: produce_item)]
     )
   end
 
--- a/spec/features/item_seasons_spec.rb
+++ b/spec/features/item_seasons_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe 'Item Seasons', type: :feature do
 
   it 'expects index to have a month' do
     visit(item_seasons_path)
-    expect(page).to have_css('div a.link_action')
+    expect(page).to have_css('div a.month_link')
   end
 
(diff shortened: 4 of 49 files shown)


vasconsaurus:
@tcione I was checking Figma and realized we had changed this page to show the months on the produce item card:
Screen Shot 2022-10-17 at 14 28 00

I'll update this PR to include this change, so there is no need to review it now. 😬

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
```
vasconsaurus, line 4: @tcione I made a helper function to help us format the month list that gets displayed inside the card. I tried to write a test, but I'm stuck. Could you give me some pointers?

I tried to use factories, but it was more confusing. I'm trying to figure out how I could write a produce_item that has many item_seasons.

app/views/produce_items/index.html.erb
```diff
@@ -1,11 +1,23 @@
...
+
+  <div class="card-holder p-top-l p-bottom-l p-sides-s">
+    <% @produce_items.find_each do |produce| %>
+      <div class="card">
+        <div class="card-inner">
+          <h4 class="produce-name"><%= I18n.t("produce_items.names.#{produce.name}") %></h4>
+          <div class="card-inner__month-name">
+            <%= produce.item_seasons.find_each do |item_season| %>
+              <%= formatted_month_list(produce, item_season) %>
+            <% end %>
```
tcione, line 16: **suggestion:** We could simplify the logic by using array methods. For example

```rb
produce.item_seasons.map { formatted_month_name(_1.month_index) }.join(' /')
```

This could be extracted to the helper if you'd like. One thing to keep in mind is that here you are relying on the order in which seasons were added to the `produce`. So `janeiro /fevereiro` and `fevereiro /janeiro` are equally valid outputs depending on the order `item_seasons` were inserted.

I would recommend making this deterministic by:

```rb
produce.item_seasons.order(month_index: :asc).map { formatted_month_name(_1.month_index) }.join(' /')
```

spec/helpers/produce_items_helper_spec.rb
tcione: Why stuck? It seems somewhat fine to me. And the usage of factories here is good, I'd do the same, since this grants you control over the data you generate (thus producing predictable results).

That said, I do have a few observations, which I'll comment in place.

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+RSpec.describe ProduceItemsHelper, type: :helper do
+  context 'when viewing the produce items index' do
+    let(:produce_item) { ProduceItem.create(name: 'carambola', category: 'fruit') }
+    let(:item_season_jan) { ItemSeason.create(produce_item_id: produce_item.id, month_index: 1, country_code: 'br') }
```
tcione, line 8: **issue:** country codes should always be uppercase as per the [ISO 3166-1 definition](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes)

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    describe '#formatted_month_name(produce, item season)' do
+      it 'adds / to the string, if its not the last or only month' do
+        expect(helper.formatted_month_list(produce_item, item_season_jan)).to eq('janeiro /')
+      end
+
+      it 'does not add / to the string, if its the last or only month' do
+        expect(helper.formatted_month_list(produce_item, item_season_fev)).to eq('fevereiro')
+      end
+    end
```
tcione, line 19: **issue:** We should shift the view here. In many ways, I don't see that much of a point of doing a helper to attach `/` or not to the end of the string. It makes a bit more sense to just go and produce the end string (in this case `janeiro /fevereiro`. That way you could shift your tests to:

1. `when produce has no seasons`
2. `when produce has one season`
3. `when produce has multiple seasons`

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+require 'rails_helper'
+
+RSpec.describe ProduceItemsHelper, type: :helper do
+  context 'when viewing the produce items index' do
+    let(:produce_item) { ProduceItem.create(name: 'carambola', category: 'fruit') }
+    let(:item_season_jan) { ItemSeason.create(produce_item_id: produce_item.id, month_index: 1, country_code: 'br') }
+    let(:item_season_fev) { ItemSeason.create(produce_item_id: produce_item.id, month_index: 2, country_code: 'br') }
+
+    describe '#formatted_month_name(produce, item season)' do
```
tcione, line 11: **issue:** no need for arguments. `#formatted_month_name` suffices

app/views/produce_items/index.html.erb
vasconsaurus: changed the helper to do this:
`produce.item_seasons.order(month_index: :asc).map { formatted_month_name(_1.month_index) }.join(' / ')`

and the index to do this:
`          <div class="card-inner__month-name">
            <%= formatted_month_list(produce) %>
          </div>`

spec/helpers/produce_items_helper_spec.rb
vasconsaurus: I have changed the helper, but I'm still not quite sure how I could write these tests

vasconsaurus:
@tcione I think we are done with this one, could you review it? 🙂

app/views/item_seasons/show.html.erb
```diff
@@ -18,7 +18,7 @@
     <% @item_season_produces.find_each do |produce| %>
       <div class="card">
         <div class="card-inner">
-          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action"  %></h4>
+          <h4 class="produce-name"><%= link_to I18n.t("produce_items.names.#{produce.name}"), produce, class: "link_action produce-name"  %></h4>
```
tcione, line 21: **question:** why does the `h4` has the same class as the link?

libpq.5.dylib
```diff
@@ -0,0 +1 @@
+/opt/homebrew/opt/postgresql@14/lib/postgresql@14/libpq.5.dylib/usr/local/lib/libpq.5.dylib
```
tcione, line 1: **question:** should this file stay in the repo?

spec/features/item_seasons_spec.rb
```diff
@@ -3,9 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe 'Item Seasons', type: :feature do
-  let(:produce_item) { ProduceItem.create(name: 'carambola', category: 'fruit') }
+  let(:produce_item) { ProduceItem.create(name: 'morango', category: 'fruit') }
```
tcione, line 6: **suggestion:** use factories instead of creating records directly

```suggestion
  let(:produce_item) do
    FactoryBot.create(
      :produce_item, 
      name: 'morango', 
      category: 'fruit',
      item_seasons: [FactoryBot.build(:item_season)]
    )
  end
```

this would allow you to remove the before block as well

spec/features/produce_items_spec.rb
```diff
@@ -3,9 +3,9 @@
 require 'rails_helper'
 
 RSpec.describe 'ProduceItems', type: :feature do
-  let!(:produce_item) { ProduceItem.create(name: 'carambola', category: 'fruit') }
+  let!(:produce_item) { ProduceItem.create(name: 'morango', category: 'fruit') }
```
tcione, line 6: **suggestion:** use FacetoryBot instead of the models directly. WDYT?

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    let!(:produce_item_two) { create(:produce_item) }
+
+    describe '#formatted_month_name' do
+      before do
+        create(:item_season, produce_item: produce_item, month_index: 1)
+        create(:item_season, produce_item: produce_item, month_index: 2)
+        create(:item_season, produce_item: produce_item_two, month_index: 3)
+      end
+
+      it 'adds / to the string, if its not the last or only month' do
```
tcione, line 17: **suggestion:** `if its not the last or only month` reads a bit weird to me. This is a bit too close to the implementation. You could say something like `formats months using / as a separator`. The fact that your expected string contains two months separated by `/` is already indication that things are working as expected. Does that make sense to you?

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe ProduceItemsHelper, type: :helper do
+  context 'when viewing the produce items index' do
+    let!(:produce_item) { create(:produce_item) }
+    let!(:produce_item_two) { create(:produce_item) }
+
+    describe '#formatted_month_name' do
```
tcione, line 10: **suggestion:** we could rethink organization a bit to read a bit better. I.e:

```rb
describe '#formatted_month_name' do
  describe 'when produce item has no seasons' do
    it 'returns an empty string'
  end

  describe 'when produce item has one season' do
    it 'returns the month name'
  end

  describe 'when produce item has multiple seasons' do
    it 'returns months separated by /'
  end
end
```

I think this brings your tests closer to the app's business language, which make them easier for humans to read and understand what you're trying to achieve. WDYT?

spec/models/item_season_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 RSpec.describe ItemSeason, type: :model do
   let!(:produce_item) { create(:produce_item) }
-  let!(:item_season) { described_class.new(produce_item_id: produce_item.id, month_index: 1, country_code: 'br') }
+  let!(:item_season) { described_class.new(produce_item_id: produce_item.id, month_index: 1, country_code: 'BR') }
```
tcione, line 7: **suggestion:** this could become `subject!(:item_season)`

tcione: **suggestion:** remove `!` from variables. I don't think we need these here.

spec/requests/produce_items_spec.rb
```diff
@@ -12,7 +12,7 @@
 
   describe 'GET /show' do
     it 'returns http success' do
-      produce_item = ProduceItem.create!(name: 'banana', category: 'fruit')
+      produce_item = ProduceItem.create!(name: 'caju', category: 'fruit')
```
tcione, line 15: **suggestion:** we could use factories instead of models directly

spec/views/item_seasons/show.html.erb_spec.rb
```diff
@@ -3,8 +3,8 @@
 require 'rails_helper'
 
 RSpec.describe 'item_seasons/show', type: :view do
-  let!(:item_season_produces) { ProduceItem.create(name: 'carambola', category: 'fruit') }
-  let!(:item_season) { ItemSeason.create(produce_item_id: item_season_produces.id, month_index: 1, country_code: 'br') }
+  let!(:item_season_produces) { ProduceItem.create(name: 'morango', category: 'fruit') }
```
tcione, line 6: **question:** why change from carambola to morango?

tcione (commented):
Nice progress <3. Left a few comments for your consideration.

app/views/item_seasons/show.html.erb
vasconsaurus: I think it is a mistake, I was probably testing what worked better and left this duplication. Looking through the files, I thought it was best to remove the class from the link. Everything still works fine.

libpq.5.dylib
vasconsaurus: I don't think so, this seems related to the Postgres path bug fix for the M1 that I had to do 😅 So I removed it from git with `git rm --cached libpq.5.dylib` and added it to the `gitignore`

spec/features/item_seasons_spec.rb
vasconsaurus: rubocop was complaining, so I changed it a bit, it seems to work

```
RSpec.describe 'Item Seasons', type: :feature do
  let(:produce_item) do
    create(
      :produce_item,
      name: 'morango',
      category: 'fruit',
      item_seasons: [build(:item_season)]
    )
  end
```

per rubocop's docs:

Screenshot 2022-11-09 at 11 56 14

spec/helpers/produce_items_helper_spec.rb
vasconsaurus: yes! completely agree 🙂

spec/helpers/produce_items_helper_spec.rb
vasconsaurus: I like this 🙂  it is indeed way easier to read. Rubocop complained about nested groups, I used `shared_context` to get around it. It works, but I'm not sure if this is the best solution.

spec/models/item_season_spec.rb
vasconsaurus: So I read a bit on subject vs. let, and it seems like it makes sense here because we are only testing for some active records validations with shoulda matchers. It is pretty clear what the subject is, and what is being tested. Is this correct?

spec/views/item_seasons/show.html.erb_spec.rb
vasconsaurus: I think I commented on this superficially, but while I was trying to write the helper test for produce_items_helper, I used before_all (I think?). And that seems to have added the fruits I tested to the DB. So now I get a `RecordNotUnique` error.

I just changed the name for a fast simple solution, but I should probably look into this. WDYT?

Screenshot 2022-11-09 at 14 51 54

spec/views/item_seasons/show.html.erb_spec.rb
```diff
@@ -3,8 +3,8 @@
 require 'rails_helper'
 
 RSpec.describe 'item_seasons/show', type: :view do
-  let!(:item_season_produces) { ProduceItem.create(name: 'morango', category: 'fruit') }
-  let!(:item_season) { ItemSeason.create(produce_item_id: item_season_produces.id, month_index: 1, country_code: 'BR') }
+  let!(:item_season_produces) { create(:produce_item, name: 'morango') }
+  let!(:item_season) { create(:item_season, produce_item: item_season_produces) }
```
vasconsaurus, line 8: @tcione I also changed this one to use factories, but I left line 13 as it was, as I'm not too sure about it. WDYT?

vasconsaurus:
@tcione There is a `NoMethodError` failure in one of our tests, but locally everything seems to be fine, those tests are passing. I tried making a couple of changes and pushing it to the repo, but I wasn't lucky 😅 Is there a better way I could debug these tests that don't fail locally?

Screenshot 2022-11-09 at 15 14 14

Screenshot 2022-11-09 at 15 16 55

spec/views/item_seasons/show.html.erb_spec.rb
vasconsaurus: So I looked around a bit, could this be a Database cleaner/Rspec configuration issue? I found this interesting, albeit old StackOverflow post: [rails-how-to-ensure-clean-state-for-test-database](https://stackoverflow.com/questions/32028909/rails-how-to-ensure-clean-state-for-test-database)

So I looked at the [database cleaner docs](https://github.com/DatabaseCleaner/database_cleaner#rspec-example), and I think this could be in the right direction. WDYT?

vasconsaurus: So I added the config here:
Screenshot 2022-11-09 at 15 49 46

Now It seems like I'm able to use 'carambola' again, AND I'm getting the failure we got in the workflows. So this seems to have fixed both those issues?

spec/features/item_seasons_spec.rb
```diff
@@ -3,14 +3,9 @@
...
-    create(
-      :produce_item,
-      name: 'morango',
-      category: 'fruit',
-      item_seasons: [create(:item_season, produce_item: produce_item)]
-    )
-  end
+  let(:produce_item) { create(:produce_item) }
+
+  before { create(:item_season, produce_item: produce_item) }
```
vasconsaurus, line 9: @tcione This doesn't seem to work. The item_season is returned(?) as nil. (Here I tried with create on line 11, but I also tried with the original build)

Question: We have the `produce_item_id` inside the `item_season`, how does the `item_seasons` array inside `produce_item` works? Looking at the schema I don't think we are able to access that information like this, but I'm not sure 😅 .

Also, while working on this, I realized we don't need to pass the name and category to `produce_item`

spec/helpers/produce_items_helper_spec.rb
tcione: Hm, tbh I kind of disagree with Rubocop's complaint, so I'd rather use my example instead of introducing a shared_context that wouldn't be really share. So, what do you think of us tweaking the rule to our liking or disabling it? :D

spec/models/item_season_spec.rb
tcione: Yeap, we should always mark what we're testing with `subject` :D

spec/views/item_seasons/show.html.erb_spec.rb
tcione: Yeap :D. That's also the correct approach when setting up tests. For some reason, I thought we were already doing that. Sorry for that.

spec/features/item_seasons_spec.rb
tcione: > Also, while working on this, I realized we don't need to pass the name and category to produce_item, our factories are doing that for us, so there is no need to overwrite them here.

Quite happy that you picked that up <3

> The old version doesn't seem to work. The item_season is returned(?) as nil. (in that version, I tried with create on line 11, but I also tried with the original build)

I lost reference to what the old version was. Would you mind pasting it here?

> Question: We have the produce_item_id inside the item_season, how does the item_seasons array inside produce_item work? Looking at the schema I don't think we are able to access that information like this, but I'm not sure 😅 .

It's a bit of FactoryBot + Rails magic. If you pass an array of `ItemSeason` to `item_seasons`, it know to:

a. Persist the produce item
b. Attach this item id to seasons

But for that to work, you need to initiatize `ItemSeason` without persisting (so, using `build` on FactoryBot and just `new` when doing pure Rails)

If you use create, it will mess up the persistance order.

spec/helpers/produce_items_helper_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        expect(helper.formatted_month_list(produce_item_one)).to eq('março')
+      end
+    end
+
+    describe 'when produce item has no seasons' do
+      it 'returns an empty string' do
+        expect(helper.formatted_month_list(produce_item_empty)).to eq('')
+      end
+    end
+  end
```
tcione, line 38: **suggestion:** we could leverage the block scopes to simplify the tests and make the tests slightly faster by doing less DB writes (writing this considering we tweaked Rubocop already):

```rb
  describe '#formatted_month_name' do
    # No need for ! here, since the variable will be used in the before blocks
    # We only need ! when we want the thing executed BEFORE its first usage
    let(:produce_item) { create(:produce_item) }

    describe 'when produce item has multiple seasons' do
      before do
        create(:item_season, produce_item: produce_item, month_index: 1)
        create(:item_season, produce_item: produce_item, month_index: 2)
      end
      
      it 'formats months using / as a separator' do
        expect(helper.formatted_month_list(produce_item)).to eq('janeiro / fevereiro')
      end
    end

    describe 'when produce item has one season' do
      before do
        create(:item_season, produce_item: produce_item, month_index: 1)
      end
      
      it 'returns the month name' do
        expect(helper.formatted_month_list(produce_item)).to eq('março')
      end
    end

    describe 'when produce item has no seasons' do
      it 'returns an empty string' do
        expect(helper.formatted_month_list(produce_item)).to eq('')
      end
    end
  end
```

tcione:
@vasconsaurus The `NoMethodError` issue seems solved. Or did I miss something?

spec/views/item_seasons/show.html.erb_spec.rb
tcione: Hm, you mean `ProduceItem.joins(:item_seasons).where(item_seasons: { month_index: item_season.month_index })`?

vasconsaurus:
> @vasconsaurus The `NoMethodError` issue seems solved. Or did I miss something?

Yes! There were two problems, that and the database cleaner config. I'm not sure they were related though. 

Anyway, I couldn't make your suggested change for the test work, so I went back to how it was.

spec/views/item_seasons/show.html.erb_spec.rb
vasconsaurus: yes 🙂

spec/features/item_seasons_spec.rb
vasconsaurus: Sorry, I think I wasn't clear, by the old version I meant your suggestion (since I changed it, now it is the old version 😅 ), this one:

```
  let(:produce_item) do
    create(
      :produce_item,
      name: 'morango',
      category: 'fruit',
      item_seasons: [create(:item_season, produce_item: produce_item)]
    )
 end
```

> If you use `create`, it will mess up the persistance order.

I used build at first, but it didn't work as well

> ... and just `new` when doing pure Rails)

What do you mean by this?

spec/helpers/produce_items_helper_spec.rb
vasconsaurus: disabled it on this test 🙂

spec/features/item_seasons_spec.rb
tcione: About using arrays: we shouldn't be setting `produce_item`, FactoryBot/Rails should that for us, so either:
```rb
  let(:produce_item) do
    create(
      :produce_item,
      name: 'morango',
      category: 'fruit',
      item_seasons: [build(:item_season, produce_item: nil)]
    )
 end
```
or 
```rb
  let(:produce_item) do
    create(
      :produce_item,
      name: 'morango',
      category: 'fruit',
      item_seasons: [build(:item_season)]
    )
 end
```
(not sure which would work).

---

> ... and just new when doing pure Rails)
>
> What do you mean by this?

`ItemSeason.new(...)`

spec/views/item_seasons/show.html.erb_spec.rb
tcione: Ah, Factories are just for creation/building. Fetching from database, update and delete should be done by normal Rails stuff.

tcione (approved):
LGTM

spec/features/item_seasons_spec.rb
vasconsaurus: Hum, I think I didn't express myself properly, I understand its `ItemSeason.new(...)`, I don't understand how everything here works together

